PR #38, opened 2025-01-17
Add NATS uri parameter for example

piotrm-nvidia:


piotrm-nvidia:
It was handled in a different way.

PR #61, opened 2025-01-24
docs: Add link to LLM example in top level README

rmccorm4:
Adds overview and links to to LLM example in top level README for discoverability.


nnshah1 (approved):
Thanks!

PR #62, opened 2025-01-24
docs: Add instructions for multi-node disaggregated deployment

ptarasiewiczNV:


examples/llm/vllm/deploy/deploy_llama_8b_disaggregated_multinode.sh
```diff
@@ -0,0 +1,141 @@
+#!/bin/bash
+
+export VLLM_ATTENTION_BACKEND=FLASHINFER
+export VLLM_WORKER_MULTIPROC_METHOD=spawn
+export VLLM_TORCH_HOST=""
+export VLLM_TORCH_PORT=36183
+export VLLM_BASELINE_WORKERS=0
+export VLLM_CONTEXT_WORKERS=1
```
nnshah1, line 8: question: for multiple context workers do we just change this variable?

nv-blazejkubiak (dismissed):
It looks pretty nice

nnshah1 (commented):
LGTM - one question: can we update the default settings in the script to be for a particular multinode scenario - say 2 x H100 x8 ? So that would be 1 gen worker TP 8 and 8 context workers TP 1?

examples/llm/vllm/README.md
```diff
@@ -319,7 +319,24 @@ In the commands above, we used the FP8 variant `neuralmagic/Meta-Llama-3.1-8B-In
...
+On the second node run the generate worker
+
+```
+./examples/llm/vllm/deploy/deploy_llama_8b_disaggregated.sh generate --head-url <head url>
+```
+
+
+
+
+## 8. Known Issues & Limitations
```
rmccorm4, line 339: Should we add a limitation on only 1 generate worker being supported at this time? (not unique to multi-node)

rmccorm4: and a limitation note on `stream: "true"` only returning 2 responses instead of N responses?

examples/llm/vllm/deploy/deploy_llama_8b_disaggregated_multinode.sh
ptarasiewiczNV: yes, this is the part of global config required by NCCL, based on that config we will calculate the world size

examples/llm/vllm/README.md
ptarasiewiczNV: Good catch! Let me add that.

ptarasiewiczNV:
> LGTM - one question: can we update the default settings in the script to be for a particular multinode scenario - say 2 x H100 x8 ? So that would be 1 gen worker TP 8 and 8 context workers TP 1?

We could, but I don't have the time to test that today. Maybe we can try this in a different PR if someone has BW for that today?

nnshah1:
> > LGTM - one question: can we update the default settings in the script to be for a particular multinode scenario - say 2 x H100 x8 ? So that would be 1 gen worker TP 8 and 8 context workers TP 1?
> 
> We could, but I don't have the time to test that today. Maybe we can try this in a different PR if someone has BW for that today?

For today let's just add a small description to the readme section detailing the setup, the script is designed to launch 1 TP 1 worker on head node and 1 TP 1 worker on second node - but can be adjusted to test different configurations.

examples/llm/vllm/README.md
```diff
@@ -319,7 +319,24 @@ In the commands above, we used the FP8 variant `neuralmagic/Meta-Llama-3.1-8B-In
...
+```
+./examples/llm/vllm/deploy/deploy_llama_8b_disaggregated.sh context --head-url <head url>
+```
+
+On the second node run the generate worker
+
+```
+./examples/llm/vllm/deploy/deploy_llama_8b_disaggregated.sh generate --head-url <head url>
+```
+
```
nnshah1, line 335:
```suggestion
The example scripts are set by default to launch one context worker with TP 1 on the head node and one generate worker with TP 1 on the secondary node. This can be changed for other configuration - see the script for details.
```

nnshah1 (approved):
LGTM - nice work! Exciting to see this.

PR #66, opened 2025-01-25
chore: Update build version to v0.1.1 and add Teardown section to LLM example README

rmccorm4:
Changes:
- Update VERSION in build.sh to reflect the latest v0.1.1 release in the image tag when building an image locally.
- Add section to README on how to cleanup processes between deployments to avoid collisions or a messy state between runs.

### DIFF
--- a/container/build.sh
+++ b/container/build.sh
@@ -17,7 +17,7 @@
 TAG=
 RUN_PREFIX=
 PLATFORM=linux/amd64
-VERSION=0.1.0
+VERSION=0.1.1
 
 # Frameworks
 #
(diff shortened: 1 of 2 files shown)


container/build.sh
```diff
@@ -17,7 +17,7 @@
 TAG=
 RUN_PREFIX=
 PLATFORM=linux/amd64
-VERSION=0.1.0
+VERSION=0.1.1
```
rmccorm4, line 20: Need to remember to update this with each release - maybe a checklist or automation

rmccorm4: Maybe some out of the box github action, or a custom one triggered on `release`: https://docs.github.com/en/actions/sharing-automations/creating-actions/releasing-and-maintaining-actions
```yaml
on:
  release:
    types: [published]
...
```

nnshah1: definitely automation - or alternate we can remove version and use git commit id or something similar

nnshah1 (approved):
LGTM - small nit: we should normalize on how we write "NOTES" - but can be done later

PR #75, opened 2025-01-30
Separate EventMetadata from Event

nv-blazejkubiak:
#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID:

- closes GitHub issue: #xxx

### DIFF
--- a/icp/src/python/triton_distributed/icp/nats_event_plane.py
+++ b/icp/src/python/triton_distributed/icp/nats_event_plane.py
@@ -136,10 +136,12 @@ class NatsEventPlane:
 
     def _extract_metadata_and_payload(self, message: bytes):
         # Extract metadata size
-        metadata_size = int.from_bytes(message[:4], byteorder="big")
+        message_view = memoryview(message)
+
+        metadata_size = int.from_bytes(message_view[:4], byteorder="big")
 
         # Extract metadata and event
-        metadata_serialized = message[4 : 4 + metadata_size]
-        event = message[4 + metadata_size :]
+        metadata_serialized = message_view[4 : 4 + metadata_size]
+        event = message_view[4 + metadata_size :]
 
         return metadata_serialized, event


icp/src/python/triton_distributed/icp/event_plane.py
```diff
@@ -45,40 +44,20 @@ def from_string(topic_str: str):
...
 
-        return Event(
-            event_id=uuid.UUID(event_pb.event_id),
-            topic=Topic.from_string(event_pb.topic),
-            event_type=event_pb.event_type,
-            timestamp=event_pb.timestamp.ToDatetime(),
-            component_id=uuid.UUID(event_pb._component_id),
-            payload=event_pb.payload,
-        )
+class EventMetadataWrapped:
```
nnshah1, line 55: could we add these to the data class?

icp/src/python/triton_distributed/icp/event_plane.py
```diff
@@ -45,40 +44,20 @@ def from_string(topic_str: str):
         return Topic(topic_str)
 
 
-@dataclass
-class Event:
-    """Event class for representing events."""
-
+class EventMetadata(BaseModel):
```
nnshah1, line 47: question for me: why base model vs dataclass?

icp/src/python/triton_distributed/icp/event_plane.py
nnshah1: is it internal? we could prefix with _

nv-blazejkubiak: I had two ideas how to make lazy deserialization - first one is with wrapper, second one is with cached properties.
Second one is cleaner from API point of view, but requires getters instead of data class.
I can switch to second option, but I will think also about the other solution.

icp/src/python/triton_distributed/icp/event_plane.py
nv-blazejkubiak: BaseModel is highly exposed in pydantic website as first choice - dataclass is more like backword compatible structure to dataclasses.dataclass (e.g. you need TypeAdapter to use some BaseModel methods on dataclass).

icp/src/python/triton_distributed/icp/nats_event_plane.py
```diff
@@ -35,44 +34,46 @@ def __init__(self, server_uri: str, component_id: uuid.UUID):
...
         event_type: Optional[str] = None,
         component_id: Optional[uuid.UUID] = None,
     ):
         async def _message_handler(msg):
-            event_pb = icp_pb2.Event()
-            event_pb.ParseFromString(msg.data)
-            event = Event.from_protobuf(event_pb)
-            await callback(event)
+            metadata, payload = self._extract_metadata_and_payload(msg.data)
+            await callback(payload, EventMetadataWrapped(metadata))
```
piotrm-nvidia, line 67: Pls add flag to submit here task without doing await here so many events can be processes concurrently.

piotrm-nvidia: Pls define callback as having optional arguments:
- payload, (mandatory)
- raw_metadata (optional)
- metadata (optional) - deserialized meta data

icp/src/python/triton_distributed/icp/nats_event_plane.py
```diff
@@ -81,3 +82,13 @@ def _comoase_subscribe_subject(
         component_id: Optional[uuid.UUID],
     ):
         return f"ep.{event_type or '*'}.{component_id or '*'}.{str(topic) + '.' if topic else ''}>"
+
+    def _extract_metadata_and_payload(self, message: bytes):
+        # Extract metadata size
+        metadata_size = int.from_bytes(message[:4], byteorder="big")
+
+        # Extract metadata and payload
+        metadata_serialized = message[4 : 4 + metadata_size]
```
piotrm-nvidia, line 91: https://docs.python.org/3/library/stdtypes.html#memoryview

Is it possible to prevent double RAM copy here with something like memory view?

icp/src/python/triton_distributed/icp/nats_event_plane.py
```diff
@@ -35,44 +34,46 @@ def __init__(self, server_uri: str, component_id: uuid.UUID):
...
         await self._nc.connect(self._server_uri)
 
-    def create_event(self, event_type: str, topic: Topic, payload: bytes):
-        event = Event(
+    async def publish(
+        self, payload: bytes, event_type: str, topic: Optional[Topic]
+    ) -> EventMetadata:
+        event_metadata = EventMetadata(
             event_id=uuid.uuid4(),
             topic=topic,
```
piotrm-nvidia, line 42: https://github.com/nats-io/nats.docs/blob/master/nats-concepts/subjects.md#characters-allowed-and-recommended-for-subject-names

There are some characters, which are forbidden in NATS.io subjects. Pls add validation.

icp/src/python/triton_distributed/icp/event_plane.py
```diff
@@ -89,17 +62,15 @@ async def connect(self):
...
+        self, payload: Union[bytes, Any], event_type: str, topic: Optional[Topic]
+    ) -> EventMetadata:
         pass
 
     @abstractmethod
     async def subscribe(
         self,
-        callback,
+        callback: Callable[[bytes, EventMetadataWrapped], NoReturn],
         topic: Optional[Topic] = None,
```
piotrm-nvidia, line 74: Does empty topic means subscribe to everything?

icp/src/python/triton_distributed/icp/event_plane.py
```diff
@@ -89,17 +62,15 @@ async def connect(self):
...
         event_type: Optional[str] = None,
```
piotrm-nvidia, line 75: If I don't set topic but set event_type, then what will be received? Is it going to receive everything and filter later based on event type? What about setting only component_id?

icp/src/python/triton_distributed/icp/event_plane.py
```diff
@@ -89,17 +62,15 @@ async def connect(self):
...
-
-    @abstractmethod
-    async def publish(self, event: Event):
+    async def publish(
+        self, payload: Union[bytes, Any], event_type: str, topic: Optional[Topic]
+    ) -> EventMetadata:
         pass
 
     @abstractmethod
     async def subscribe(
```
piotrm-nvidia, line 71: Pls add docstring with examples.

icp/src/python/triton_distributed/icp/event_plane.py
```diff
@@ -89,17 +62,15 @@ async def connect(self):
         pass
 
     @abstractmethod
-    def create_event(self, event_type: str, topic: Topic, payload: bytes):
-        pass
-
-    @abstractmethod
-    async def publish(self, event: Event):
+    async def publish(
+        self, payload: Union[bytes, Any], event_type: str, topic: Optional[Topic]
```
piotrm-nvidia, line 66: Why there is not component id here but it is in subscribe?

piotrm-nvidia: Can I publish something without component id?

icp/src/python/triton_distributed/icp/event_plane.py
```diff
@@ -89,17 +62,15 @@ async def connect(self):
         pass
 
     @abstractmethod
-    def create_event(self, event_type: str, topic: Topic, payload: bytes):
-        pass
-
-    @abstractmethod
-    async def publish(self, event: Event):
+    async def publish(
```
piotrm-nvidia, line 65: Pls add doctoring with examples of publish calls.

icp/src/python/triton_distributed/icp/nats_event_plane.py
```diff
@@ -35,44 +34,46 @@ def __init__(self, server_uri: str, component_id: uuid.UUID):
     async def connect(self):
         await self._nc.connect(self._server_uri)
 
-    def create_event(self, event_type: str, topic: Topic, payload: bytes):
-        event = Event(
+    async def publish(
+        self, payload: bytes, event_type: str, topic: Optional[Topic]
+    ) -> EventMetadata:
+        event_metadata = EventMetadata(
             event_id=uuid.uuid4(),
```
piotrm-nvidia, line 41: Pls consider optional event_id parameter so I can set my own event id.

icp/src/python/triton_distributed/icp/nats_event_plane.py
```diff
@@ -35,44 +34,46 @@ def __init__(self, server_uri: str, component_id: uuid.UUID):
...
             topic=topic,
             event_type=event_type,
             timestamp=datetime.utcnow(),
```
piotrm-nvidia, line 44: It is possible that even happened before this call. Pls add optional timestamp.

icp/src/python/triton_distributed/icp/event_plane.py
nv-blazejkubiak: yes. there is root topic

icp/src/python/triton_distributed/icp/event_plane.py
nv-blazejkubiak: component_id is always added by event plane

icp/src/python/triton_distributed/icp/nats_event_plane.py
nv-blazejkubiak: this would complicate API - let's skip this for now

icp/src/python/triton_distributed/icp/nats_event_plane.py
nv-blazejkubiak: this would complicate API - let's skip this for now

icp/src/python/triton_distributed/icp/nats_event_plane.py
```diff
@@ -35,44 +34,46 @@ def __init__(self, server_uri: str, component_id: uuid.UUID):
...
         subject = self._comoase_subscribe_subject(topic, event_type, component_id)
         await self._nc.subscribe(subject, cb=_message_handler)
 
     async def disconnect(self):
         await self._nc.close()
 
-    def _compose_publish_subject(self, event: Event):
-        return f"ep.{event.event_type}.{event.component_id}.{event.topic}.trunk"
+    def _compose_publish_subject(self, event_metadata: EventMetadata):
+        return f"ep.{event_metadata.event_type}.{event_metadata.component_id}.{str(event_metadata.topic) + '.' if event_metadata.topic else ''}trunk"
```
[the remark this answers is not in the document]
nv-blazejkubiak: It is not

icp/src/python/triton_distributed/icp/event_plane.py
nv-blazejkubiak: then you get all events from all topics from that component

icp/src/python/triton_distributed/icp/event_plane.py
nnshah1: We also got feedback though that pydantic types can be heavy and that in some cases data classes are better, 

https://ai.pydantic.dev/#hello-world-example

If you look in the hello world - they use both 

so want to make sure in this case we need the pydantic base model support - or can use the simpler dataclass

If you have a reference on when to use what -would like to learn.

icp/src/python/triton_distributed/icp/event_plane.py
```diff
@@ -14,98 +14,125 @@
...
+
+    :param subjects: List of strings to validate
+    :return: True if all strings are valid, False otherwise
+    """
+    pattern = re.compile(r"^[a-zA-Z0-9_-]+$")
+
+    return all(pattern.match(subject) for subject in subjects)
+
+
+class Topic(BaseModel):
```
nnshah1, line 38: for this in particular seems overkill to be from basemodel?

icp/src/python/triton_distributed/icp/event_plane.py
```diff
@@ -14,98 +14,125 @@
...
-            topic=Topic.from_string(event_pb.topic),
-            event_type=event_pb.event_type,
-            timestamp=event_pb.timestamp.ToDatetime(),
-            component_id=uuid.UUID(event_pb._component_id),
-            payload=event_pb.payload,
-        )
+
+    @classmethod
+    def from_raw(cls, event_metadata_serialized: bytes):
+        return cls.model_validate_json(event_metadata_serialized)
```
nnshah1, line 76: this forces us to use json for the event metadata? I think maybe protobuf or flatbuffers may be better?

icp/src/python/triton_distributed/icp/event_plane.py
```diff
@@ -14,98 +14,125 @@
...
+            callback (Callable[[bytes, bytes], Awaitable[None]]): Callback function to be called when an event is received
+            topic (Optional[Topic]): Event topic
+            event_type (Optional[str]): Event type
+            component_id (Optional[uuid.UUID]): Component ID
+        """
         pass
 
     @abstractmethod
-    async def subscribe(
+    def subscribe_iter(
```
nnshah1, line 119: can we do one with args to control?

PR #84, opened 2025-02-03
build: independent pip

nnshah1:
#### What does the PR do?

Add wheels and make tritonserver optional

#### Checklist

#### Commit Type:
- [x] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [x] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID:

- closes GitHub issue: #xxx


icp/src/python/triton_distributed/icp/data_plane.py
```diff
@@ -73,13 +57,13 @@ def set_icp_data_type(
...
 ) -> None:
-    message.datatype = DataTypeString(value)
+    message.datatype = value.name
 
 
 def get_icp_data_type(
     message: ModelInferRequest.InferInputTensor | ModelInferResponse.InferOutputTensor,
 ) -> DataType:
-    return StringToDataType(message.datatype)
+    return DataType(message.datatype)
```
nnshah1, line 66: todo: fix - this doesn't work

icp/src/python/triton_distributed/icp/data_plane.py
```diff
@@ -116,17 +100,15 @@ def set_icp_memory_type(
...
 
 def get_icp_memory_type(
     message: ModelInferRequest.InferInputTensor | ModelInferResponse.InferOutputTensor,
 ) -> MemoryType | None:
     if ICP_MEMORY_TYPE not in message.parameters:
         return None
-    return STRING_TO_TRITON_MEMORY_TYPE[
-        message.parameters[ICP_MEMORY_TYPE].string_param
-    ]
+    return MemoryType(message.parameters[ICP_MEMORY_TYPE].string_param)
```
nnshah1, line 111: todo: fix

icp/src/python/triton_distributed/icp/data_type.py
```diff
@@ -0,0 +1,89 @@
+# Copyright 2023-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
nnshah1, line 1: upate license

runtime/python/src/triton_distributed/runtime/triton_core_operator.py
```diff
@@ -90,13 +106,87 @@ def __init__(
...
+                tensor.memory_buffer.data_ptr,
+                TritonMemoryType(tensor.memory_buffer.memory_type),
+                tensor.memory_buffer.memory_type_id,
+                tensor.memory_buffer.size,
+                tensor.memory_buffer.owner,
+            ),
+        )
+
+    @staticmethod
+    def _remote_request_to_local_request(
```
tanmayv25, line 124: `_to_triton_core_request` ?
The input arg is already  described as RemoteInferenceRequest.
If still needed, how about `_remote_to_triton_core_request`?

runtime/python/src/triton_distributed/runtime/triton_core_operator.py
```diff
@@ -90,13 +106,87 @@ def __init__(
...
+        request: RemoteInferenceRequest, model: tritonserver.Model
+    ):
```
tanmayv25, line 126: Why not add the return type with this function? `-> InferenceRequest`? Where InferenceRequest comes from tritonserver._api._request

runtime/python/src/triton_distributed/runtime/triton_core_operator.py
```diff
@@ -90,13 +106,87 @@ def __init__(
             model_config = None
         self._local_model = self._triton_core.load(self._name, model_config)
 
+    @staticmethod
+    def _to_triton_tensor(tensor: Tensor):
```
tanmayv25, line 110: return type hints as well.

runtime/python/src/triton_distributed/runtime/triton_core_operator.py
```diff
@@ -90,13 +106,87 @@ def __init__(
...
+
+    @staticmethod
+    def _set_local_request_parameters(
+        request: RemoteInferenceRequest, local_request: tritonserver.InferenceRequest
+    ):
+        for parameter_name, parameter_value in request.parameters.items():
+            local_request.parameters[parameter_name] = parameter_value
+
+    @staticmethod
+    def _local_response_to_remote_response(
```
tanmayv25, line 159: `_from_triton_core_response`? Use type hints to describe the flow.

container/Dockerfile
```diff
@@ -154,8 +154,13 @@ COPY . /workspace
 RUN cd runtime/rust && cargo build --release --locked && cargo doc --no-deps
 RUN /workspace/icp/protos/gen_python.sh
 
+# Install python packages
+ARG PYTHON_PACKAGE_VERSION
+RUN SETUPTOOLS_SCM_PRETEND_VERSION_FOR_TRITON_DISTRIBUTED_ICP=${PYTHON_PACKAGE_VERSION} pip install -e /workspace/icp/python
+RUN SETUPTOOLS_SCM_PRETEND_VERSION_FOR_TRITON_DISTRIBUTED_RUNTIME=${PYTHON_PACKAGE_VERSION} pip install -e /workspace/runtime/python
```
rmccorm4, line 160: I had to do some googling about these - so added some comments so the next reader might not need to google. Feel free to edit the wording
```suggestion
# SETUPTOOLS_SCM_PRETEND_VERSION allows dynamically setting the package versions during build/install.
# This allows having versioned packages during development between releases, such as commit IDs.
RUN SETUPTOOLS_SCM_PRETEND_VERSION_FOR_TRITON_DISTRIBUTED_ICP=${PYTHON_PACKAGE_VERSION} pip install -e /workspace/icp/python
RUN SETUPTOOLS_SCM_PRETEND_VERSION_FOR_TRITON_DISTRIBUTED_RUNTIME=${PYTHON_PACKAGE_VERSION} pip install -e /workspace/runtime/python
```

runtime/python/src/triton_distributed/runtime/triton_core_operator.py
```diff
@@ -90,13 +105,87 @@ def __init__(
             model_config = None
         self._local_model = self._triton_core.load(self._name, model_config)
```
tanmayv25, line 106: `_local_model` -> `_triton_core_model`?

icp/python/src/triton_distributed/icp/tensor.py
tanmayv25, line 1: This is a copy from here right? 
https://github.com/triton-inference-server/core/blob/main/python/tritonserver/_api/_tensor.py

I believe you copied it because you wanted to decouple the tritonserver package from usual triton_distributed project installation. Specifically, if the tritonserver package is not installed, then still be able to use triton-distributed. However, this code duplication can be difficult to maintain in long-term. 
What are your long-term plans?

tanmayv25: Similar comment for other newly added files which appear to be replicated code between the projects.

container/Dockerfile
```diff
@@ -154,8 +154,20 @@ COPY . /workspace
 RUN cd runtime/rust && cargo build --release --locked && cargo doc --no-deps
 RUN /workspace/icp/protos/gen_python.sh
 
+# Install python packages
+ARG PYTHON_PACKAGE_VERSION
```
tanmayv25, line 158: Should we add a default version with the repo which is updated with every release?
What happens when  version is not provided?

icp/python/src/triton_distributed/icp/tensor.py
nnshah1: goal is to keep them compatible - but independent. 

They have the same API - as we evolve to support non tensors - will need to see if its still needed. 

The tensors will continue to support dlpack and in some sense that is the main interface between the two.

container/Dockerfile
nnshah1: this is passed in via build.sh
build.sh now auto sets this based on git -
it finds the latest tag, if it matches that is used, if not it adds the commit_id and the term 'dev' after.
can add a default here though - just in case

tanmayv25 (approved):

PR #92, opened 2025-02-04
Normalize NATS.io host and port

piotrm-nvidia:


ptarasiewiczNV (approved):
great, thanks for that!

ryanolson:
Our current configs are a bit scattered with the move, but there is a great crate called figment which can be used to parse ENVs, toml, json, yaml, etc files or any serde object.

Rust builders are also serde objects and so are clap/cli options.

We could use a single config/settings structure to handling extracting the config from the runtime system's state + cli + derived overrides.

This shows a partial:
https://github.com/triton-inference-server/triton_distributed/blob/main/runtime/rust/src/config.rs#L89

Simple object, trivial to make pybinds.

examples/python/hello_world/client/parser.py
```diff
@@ -15,14 +15,16 @@
 
 import argparse
 
+from triton_distributed.worker.worker import DEFAULT_REQUESTS_URI
```
glos-nv, line 18: Is this variable really defined there?

glos-nv (changes requested):
Tests are failing due to incorrect imports.

examples/python/llm/vllm/deploy/parser.py
```diff
@@ -1,5 +1,7 @@
 import argparse
 
+from triton_distributed.worker.worker import DEFAULT_REQUESTS_URI
```
glos-nv, line 3: This import is probably still wrong

examples/python/hello_world/client/parser.py
piotrm-nvidia: Ruff removed this from worker and also this code is outdated because worker changes to runtime.

glos-nv (dismissed):
Now it LGTM ;)

container/Dockerfile
```diff
@@ -122,6 +122,10 @@ ENV PYTHONUNBUFFERED=1
 # Install NATS - pointing toward NATS github instead of binaries.nats.dev due to server instability
 RUN wget https://github.com/nats-io/nats-server/releases/download/v2.10.24/nats-server-v2.10.24-amd64.deb && dpkg -i nats-server-v2.10.24-amd64.deb
 
+ENV DEFAULT_REQUESTS_PORT=4222
```
whoisj, line 125: any chance we properly namespace these ENV names?

something like 
```
TRITON_REQUESTS_PORT
TRITON_REQUESTS_HOST
TRITON_REQUESTS_URI
```
could work, maybe?

in the Helm charts, I used the following (all placeholder for now):

```
TRITON_REQUEST_PLANE_HOST
TRITON_REQUEST_PLANE_PORT
```
as well as
```
TRITON_LLM_PP
TRITON_LLM_TP
TRITON_PORT_HEALTH
TRITON_PORT_METRICS
TRITON_PORT_REQUESTS
```
for the Triton specific environment variables to avoid any conflict with anything else consumers of our libraries might want to use.

icp/python/src/triton_distributed/icp/nats_request_plane.py
```diff
@@ -39,6 +39,12 @@
     set_icp_response_to_uri,
 )
 
+DEFAULT_REQUESTS_PORT = int(os.getenv("DEFAULT_REQUESTS_PORT", 4222))
+DEFAULT_REQUESTS_HOST = os.getenv("DEFAULT_REQUESTS_HOST", "localhost")
+DEFAULT_REQUESTS_URI = os.getenv(
+    "DEFAULT_REQUESTS_URI", f"nats://{DEFAULT_REQUESTS_HOST}:{DEFAULT_REQUESTS_PORT}"
```
whoisj, line 45: while we're in here making things better, can we PLEASE PLEASE PLEASE not hard code `nats:` as the URL protocol?

> TLS encrypted only TCP connection (i.e. NATS URLs starting with `tls://...`)
>
> TLS encrypted if the server is configured for it or plain un-encrypted TCP connection otherwise (i.e. NATS URLs starting with `nats://...`)

given that use of `nats://` forced unencrypted communication, having it be a variable seems like the right thing to do here.

icp/python/src/triton_distributed/icp/nats_request_plane.py
```diff
@@ -94,10 +108,12 @@ def __init__(
     ) -> None:
         self._process = None
         self.port = port
-        self.url = f"nats://localhost:{port}"
+        self.url = f"{host}:{port}"
```
whoisj, line 111: does `host` include the URL protocol, if not then we likely need that prepended here.

icp/python/src/triton_distributed/icp/nats_request_plane.py
```diff
@@ -159,7 +181,7 @@ def __del__(self):
 
     def __init__(
         self,
-        request_plane_uri: str = "nats://localhost:4222",
+        request_plane_uri: str = DEFAULT_REQUESTS_URI,
```
whoisj, line 184: maybe we just need a static global function that composes and returns the correct URL for nats?

whoisj:
please add a description to your pull-request as that's what will be used when the squashed commit is generated (🤬)

rmccorm4:
Do we still need this PR or can it be closed for now?

piotrm-nvidia:
It won't be implemented.

PR #94, opened 2025-02-04
feat: add python bindings + wheel build

ryanolson:
#### What does the PR do?

Python Bindings + UV/Maturin Wheel build

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [x] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:
@nnshah1 

examples, then rust/lib.rs and rust/engine.rs

#### Test plan:

build wheel, install, rust hello_world.py for integration test

- CI Pipeline ID:

@nnshah1


runtime/rust/python-wheel/.gitignore
```diff
@@ -0,0 +1,2 @@
+/target
+python/nova_distributed/*.so
```
nnshah1, line 2:
```suggestion
python/triton_distributed/*.so
```

runtime/rust/python-wheel/Cargo.lock
```diff
@@ -0,0 +1,3015 @@
+# This file is automatically @generated by Cargo.
+# It is not intended for manual editing.
+version = 4
+
+[[package]]
+name = "addr2line"
```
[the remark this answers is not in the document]
ryanolson: you can use `cargo license` to dump a text file or a jsonl file, then use a jinja template to render it.

```
cargo install cargo-license
cargo license
```

There should be a CI check with a license.lock file to ensure that if you regenerate the license file it's idempotent.  if it changes the file o rmakes the git dirty, we fail CI.  

One or more OSRB reps should be assigned as codeowner on the license.lock file

This will ensure that we don't bring in any new packages that are not approved.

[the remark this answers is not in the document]
ryanolson: @nnshah1 ☝️

runtime/rust/python-wheel/examples/bls/bls.py
nnshah1, line 1: could we rename this to something - maybe pipeline or chain?

want to avoid using the term bls - and reserve that for triton core model - plus BLS to me is not different than any other asyncengine / operator -

runtime/rust/python-wheel/.gitignore
```diff
@@ -0,0 +1,2 @@
+/target
+python/triton_distributed/*.so
```
nnshah1, line 2: Awesome

runtime/rust/python-wheel/Cargo.toml
```diff
@@ -0,0 +1,41 @@
+[package]
+name = "nova_distributed_py3"
```
nnshah1, line 2:
```suggestion
name = "triton_distributed_py3"
```

runtime/rust/python-wheel/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+
+[lib]
+path = "rust/lib.rs"
+name = "_core"
+# "cdylib" is necessary to produce a shared library for Python to import from.
+crate-type = ["cdylib"]
+
+
+[dependencies]
+triton-distributed = { path = "../" }
```
nnshah1, line 14: Does this need to be snake case?
Could it be?

runtime/rust/python-wheel/examples/bls/bar.py
```diff
@@ -0,0 +1,25 @@
+import asyncio
+
+import uvloop
+from nova_distributed import DistributedRuntime, nova_worker
```
nnshah1, line 4: will probably need a global replace
```suggestion
from triton_distributed import DistributedRuntime, nova_worker
```

runtime/rust/python-wheel/examples/bls/bls.py
ryanolson: where is already an example of called pipeline.  

the pipeline chaining request transformations, then chain response transformations.

the example currently named bls, is more of a composition example, the output of one call are chained as inputs to a second call.

runtime/rust/python-wheel/Cargo.toml
ryanolson: rust packages are not exclusive `-` delimited, but the best ones are.  when i look down my dependencies list the majority are `-` delimited, only the odd ducks are `_` delimited. when it doubt, `-` it out...

runtime/rust/python-wheel/examples/bls/bar.py
ryanolson: updated all `nova_` and `nova-`

ptarasiewiczNV (commented):
Please rename `runtime/rust/python-wheel/python/nova_distributed` -> `runtime/rust/python-wheel/python/triton_distributed`
With that change I was able to build the wheel and run the hello world example.

runtime/rust/python-wheel/Cargo.toml
nnshah1: ok - otherwise we'll some places with "_" and some with "-"

ok with that-if that is the convention

runtime/rust/python-wheel/examples/bls/bar.py
nnshah1, line 1: need copyright

runtime/rust/python-wheel/Cargo.lock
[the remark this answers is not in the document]
dmitry-tokarev-nv: Right now I run a python script to do that locally. Will work PR check this in near term future.

PR #96, opened 2025-02-04
ci: restrict which branches & paths trigging lab testing

whoisj:
This change adds filters to both pull and push automation scripts that limit when work will sent to the NVIDIA Test Lab based on the target branch of and paths modified by a pull request.

This should help prevent unnecessary work from being sent to lab machines.

### DIFF
--- a/.github/workflows/trigger_ci_pull.yml
+++ b/.github/workflows/trigger_ci_pull.yml
@@ -13,6 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+name: NVIDIA Test Lab (pull) Validation
+
 on:
   pull_request:
     paths-ignore:
--- a/.github/workflows/trigger_ci_push.yml
+++ b/.github/workflows/trigger_ci_push.yml
@@ -13,13 +13,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+name: NVIDIA Test Lab (push) Validation
+
 on:
   push:
     branches:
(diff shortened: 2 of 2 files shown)


.github/workflows/trigger_ci_pull.yml
```diff
@@ -15,23 +15,27 @@
 
 on:
   pull_request:
+    branches:
```
rmccorm4, line 18: @nv-anants can you TLDR; the pull vs push flows here? Was `pull` intended to happen on any pull request as previously defined?

rmccorm4 (commented):
`trigger_ci_push.yml` LGTM - left a question on `trigger_ci_pull.yml`

.github/workflows/trigger_ci_pull.yml
whoisj: according to the documentation the value of `branch` here matches against the target branch, and not the PR's branch.

nv-anants: restricting to pull requests on main is fine. Original intention was to keep it open i believe, but i dont see much use for it today. So after this change -  
we'll run gitlab CI on main branch commits and pull requests to main only

rmccorm4 (approved):

nv-anants (approved):

whoisj:
forced to rebase because something merged into `main`. It would be really nice if we actually knew how to use Git properly. 😭

rmccorm4 (approved):

PR #109, opened 2025-02-05
Add section in contributions about CI.

piotrm-nvidia:


CONTRIBUTING.md
nnshah1, line 1: I think we should create a separate readme around testing - let us add those there

PR #118, opened 2025-02-06
fix: Set default 'latest' tag if git tags fail to get parsed

rmccorm4:
Adds fallback logic to set `latest_tag=0.0.1` if git tags aren't found for whatever reason, which would previously lead to an error with the SCM version during pip install:
```
fatal: No names found, cannot describe anything.
...

PYTHON_PACKAGE_VERSION=.dev+6c62884

...

packaging.version.InvalidVersion: Invalid version: '.dev+6c62884'
```


nnshah1 (approved):
thanks for the quick fix

PR #121, opened 2025-02-06
feat: TritonCore DistributedRuntime Worker Example

rmccorm4:


runtime/rust/python-wheel/examples/triton_core/server.py
```diff
@@ -0,0 +1,138 @@
...
+        if not isinstance(request, str):
+            raise ValueError("Request must be a string")
+
+        try:
+            print(f"Processing generation request: {request}")
+            text_input: List[str] = [request]
+            stream: List[bool] = [True]
+
+            triton_core_inputs: Dict[str, Any] = {
+                "text_input": text_input,
```
nnshah1, line 96: nice!

PR #122, opened 2025-02-06
ci: Remove conflicting paths-ignore from rust action

rmccorm4:
Fixes failures like this: https://github.com/triton-inference-server/triton_distributed/actions/runs/13171148554/workflow

``` 
GitHub Actions
/ .github/workflows/pre-merge-rust.yml
Invalid workflow file

you may only define one of `paths` and `paths-ignore` for a single event
```

Shouldn't need to exclude kubernetes directory changes here since it is already specific to changes made in rust directory

### DIFF
--- a/.github/workflows/pre-merge-rust.yml
+++ b/.github/workflows/pre-merge-rust.yml
@@ -27,8 +27,6 @@ on:
     - main
     paths:
     - 'runtime/rust/**'
-    paths-ignore:
-    - deploy/Kubernetes/**
 
 jobs:
   pre-merge-rust:


.github/workflows/pre-merge-rust.yml
```diff
@@ -27,8 +27,6 @@ on:
     - main
     paths:
     - 'runtime/rust/**'
-    paths-ignore:
```
nv-anants, line 30: adding `- ! deploy/Kubernetes/**` under path should work, but i don't think there's a need for it

nv-anants (approved):

.github/workflows/pre-merge-rust.yml
rmccorm4: Yep, I did that originally then realized it wasn't necessary

PR #132, opened 2025-02-08
refactor: Modify rust vllm example to use container

rmccorm4:


nnshah1 (approved):

rmccorm4:
virtual environment flow has its pros and cons. It can be a bit tedious - and can't easily a pytest that uses `triton_distributed_rs` without changing the flow to activate the virtual environment first.

However, installing vllm directly in the container without a virtual environment ran into some issues described here for follow-up: https://github.com/triton-inference-server/triton_distributed/issues/134.

Also note - the `triton_distributed_rs` doens't work with `./container/run.sh --mount-workspace` because `maturin develop` creates an editable install of the wheel, but when mounting in the code, it overwrites the location of the wheel from host where it may not have been built yet - and corrupts the install.

This can be fixed by using `maturin build` instead of `maturin develop` to get a non-editable install. See https://github.com/triton-inference-server/triton_distributed/pull/135

PR #152, opened 2025-02-10
ci: Move slow pytests to nightly

rmccorm4:
Marks slower mock_disagg and perf_benchmark tests to nightly instead of pre_merge.

This should cut down pytest time from ~5-10min to ~1-2min.

Corresponding gitlab PR to run nightly tests coming soon.

### DIFF
--- a/runtime/tests/python/integration/test_mock_disaggregated_serving.py
+++ b/runtime/tests/python/integration/test_mock_disaggregated_serving.py
@@ -58,10 +58,8 @@ try:
 except CUDARuntimeError:
     print("CUDA not available")
 
-# TODO
-# Decide if this should be
-# pre merge, nightly, or weekly
-pytestmark = pytest.mark.pre_merge
+# Slower test than others - make it nightly for now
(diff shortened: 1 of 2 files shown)


runtime/tests/python/integration/test_mock_disaggregated_serving.py
```diff
@@ -58,10 +58,8 @@
 except CUDARuntimeError:
     print("CUDA not available")
 
-# TODO
-# Decide if this should be
-# pre merge, nightly, or weekly
-pytestmark = pytest.mark.pre_merge
+# Slower test than others - make it nightly for now
+pytestmark = pytest.mark.nightly
```
nnshah1, line 62: one quick note -I think we want to list out the test marks in the conftest file I believe or project.toml

nnshah1: I see already done :)

nnshah1 (approved):
LGTM

PR #153, opened 2025-02-10
ci: add path checks for vllm framework testing

nv-anants:
#### What does the PR do?
* Merge trigger_ci_push and pull. Both use same workflow, except the reference passed in pipeline trigger API. Using if/else now
* Check for changes in vllm associated path and use variable to run vllm test only when changes are detected

vllm changes add variable to pipeline trigger request - https://github.com/triton-inference-server/triton_distributed/actions/runs/13272122711/job/37053800450
```
Running Pipeline with Variables: --form variables[RUN_VLLM]=true 
```
Pipeline - https://gitlab-master.nvidia.com/dl/triton/triton-distributed/-/pipelines/23807404

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID:

- closes GitHub issue: #xxx


.github/workflows/trigger_ci_pull.yml
```diff
@@ -37,7 +36,23 @@ jobs:
     needs: mirror_repo
     runs-on: self-hosted
     steps:
+    - name: Detect source code changes
+      uses: dorny/paths-filter@v3
+      id: src
+      with:
+        filters: |
+          vllm:
+            - 'examples/python/llm/vllm/**'
```
rmccorm4, line 45: We should probably also add these paths:
- `container/deps/vllm` directory
- `container/deps/requirements.vllm.txt` file

rmccorm4, line 45: double check my syntax here - but let's move it one directory up because the `llm/api_server` dir has some shared implementation the `vllm` also uses
```suggestion
            - 'examples/python/llm/**'
```

.github/workflows/trigger_ci.yml
```diff
@@ -0,0 +1,82 @@
...
+    - name: Detect source code changes
+      id: src_changes
+      uses: dorny/paths-filter@v3
+      with:
+        filters: |
+          vllm:
+            - 'examples/python/llm/**'
+            - 'container/deps/requirements.vllm.txt'
+            - 'container/deps/vllm/**'
+    - name: Trigger Pipeline
```
rmccorm4, line 57: I think at some point soon it would be more clear to end users and developers if we had separate line-items reported back to github:
- ✅ standard pipeline
- ✅ vllm pipeline

such that when a vllm pipeline is _not_ triggered, it would be clearly evident in the PR to see
- ✅ standard pipeline
- `skipped` vllm pipeline

I think this will help us catch logical errors faster and notice when we are not triggering certain pipelines when we should be

rmccorm4 (approved):
LGTM, thanks @nv-anants! 

I think we'll want to make these separate logical jobs/checks when reported back to github in the near future to avoid accidentally skipping pipelines when developers aren't aware their expected tests aren't being run

https://github.com/triton-inference-server/triton_distributed/pull/153/files#r1951756750

PR #156, opened 2025-02-11
refactor: Handle JSON serialize / de-serialize errors.

grahamking:
Replace `unwrap()` with error handling at the interface between JSON and Rust; where weak typing becomes strong typing.

### DIFF
--- a/runtime/rust/python-wheel/rust/lib.rs
+++ b/runtime/rust/python-wheel/rust/lib.rs
@@ -338,7 +338,16 @@ async fn process_stream(
     let mut stream = stream;
     while let Some(response) = stream.next().await {
         // Convert the response to a PyObject using Python's GIL
-        let annotated: RsAnnotated<serde_json::Value> = serde_json::from_value(response).unwrap();
+        // TODO: Remove the clone, but still log the full JSON string on error. But how?
+        let annotated: RsAnnotated<serde_json::Value> = match serde_json::from_value(
+            response.clone(),
+        ) {
+            Ok(a) => a,
+            Err(err) => {
+                tracing::error!(%err, %response, "process_stream: Failed de-serializing JSON into RsAnnotated");
+                break;
+            }
+        };
 
         let annotated: RsAnnotated<PyObject> = annotated.map_data(|data| {
             let result = Python::with_gil(|py| match pythonize::pythonize(py, &data) {
--- a/runtime/rust/src/pipeline/network/tcp/client.rs
+++ b/runtime/rust/src/pipeline/network/tcp/client.rs
@@ -16,6 +16,7 @@
 use std::sync::Arc;
 
 use futures::{SinkExt, StreamExt};
+use tokio::io::{ReadHalf, WriteHalf};
 use tokio::{io::AsyncWriteExt, net::TcpStream};
 use tokio_util::codec::{FramedRead, FramedWrite};
 
@@ -82,7 +83,7 @@ impl TcpClient {
         let stream = TcpClient::connect(&info.address).await?;
         let (read_half, write_half) = tokio::io::split(stream);
 
-        let mut framed_reader = FramedRead::new(read_half, TwoPartCodec::default());
+        let framed_reader = FramedRead::new(read_half, TwoPartCodec::default());
         let mut framed_writer = FramedWrite::new(write_half, TwoPartCodec::default());
 
         // this is a oneshot channel that will be used to signal when the stream is closed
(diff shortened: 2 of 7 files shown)


runtime/rust/python-wheel/rust/lib.rs
```diff
@@ -339,7 +339,16 @@ async fn process_stream(
     let mut stream = stream;
     while let Some(response) = stream.next().await {
         // Convert the response to a PyObject using Python's GIL
-        let annotated: RsAnnotated<serde_json::Value> = serde_json::from_value(response).unwrap();
+        // TODO: Remove the clone, but still log the full JSON string on error. But how?
+        let annotated: RsAnnotated<serde_json::Value> = match serde_json::from_value(
+            response.clone(),
```
rmccorm4, line 344: Is this adding a copy on every response? This is a nice example of where having automated benchmark/dashboard in place would be great to help affirm these types of changes or detect regressions, even if only on a dummy engine/worker - CC @kthui @glos-nv

rmccorm4: @grahamking do you have any manual kind of benchmark that could quickly show the impact? For example on a large output sequence length?

runtime/rust/python-wheel/rust/lib.rs
```diff
@@ -339,7 +339,16 @@ async fn process_stream(
     let mut stream = stream;
     while let Some(response) = stream.next().await {
         // Convert the response to a PyObject using Python's GIL
-        let annotated: RsAnnotated<serde_json::Value> = serde_json::from_value(response).unwrap();
```
rmccorm4, line 342: For my own understanding - is the move from `unwrap()` to `Ok/Err` handling to log nice errors instead of `panic`ing on an error?

runtime/rust/python-wheel/rust/lib.rs
```diff
@@ -339,7 +339,16 @@ async fn process_stream(
...
+        // TODO: Remove the clone, but still log the full JSON string on error. But how?
+        let annotated: RsAnnotated<serde_json::Value> = match serde_json::from_value(
+            response.clone(),
+        ) {
+            Ok(a) => a,
+            Err(err) => {
+                log::error!(%err, %response, "process_stream: Failed de-serializing JSON into RsAnnotated");
+                break;
+            }
+        };
```
rmccorm4, line 351: I believe if we remove `%response` from the log, we can avoid the `clone()` - but I'm not sure how much this hurts the logging clarity.
```suggestion
        let annotated: RsAnnotated<serde_json::Value> = match serde_json::from_value(
            response,
        ) {
            Ok(a) => a,
            Err(err) => {
                log::error!(%err, "process_stream: Failed de-serializing JSON into RsAnnotated");
                break;
            }
        };
```

rmccorm4, line 351: I believe if we remove %response from the log, we can avoid the clone() - but I'm not sure how much this hurts the logging clarity.

```suggestion
        let annotated: RsAnnotated<serde_json::Value> = match serde_json::from_value(response) {
            Ok(a) => a,
            Err(err) => {
                log::error!(%err, "process_stream: Failed de-serializing JSON into RsAnnotated");
                break;
            }
        };
```

runtime/rust/python-wheel/rust/lib.rs
grahamking: Exactly. A `panic!` will kill the process. This way we only stop the errant request.

grahamking:
@rmccorm4 About the clone

The error `serde` (the serialize/de-serialize library) gives us if a de-serialize fails look something like this:
> Error at line 1 column 253 expected '}'

Without the JSON it's useless.

Here's what I propose:
- We leave it in while we are in the "make it right" phase of the project. We will take it out during the "make it fast" phase. It's almost impossible to debug JSON errors without knowing the JSON.
- We write some benchmarks. I wager they will be completely dominated by the Python part, which is all heap allocated, and this single allocation/copy will be invisible. We'll still remove it later though.

runtime/rust/python-wheel/rust/lib.rs
grahamking: No, not yet. Strong +1 on us getting that though.

Mostly malloc should be able to re-use memory from the previous iteration, so it's not talking to the kernel. It will very likely zero it (`memset`) each time.

If `RsAnnotated` is over 128K (default) this becomes `mmap` / `munmap` each time, but I don't think we'd be in that case.

I wager this will be invisible on performance profiles though, there are many other allocations and copies in the critical path.

runtime/rust/src/pipeline/network/tcp/client.rs
```diff
@@ -189,3 +137,80 @@ impl TcpClient {
...
+                           (Some(bytes), None) => {
+                                let msg: ControlMessage = match serde_json::from_slice(bytes) {
+                                    Ok(msg) => msg,
+                                    Err(err) => {
+                                        let json_str = String::from_utf8_lossy(bytes);
+                                        log::error!(%err, %json_str, "control_handler fatal error deserializing JSON to ControlMessage");
+                                        context.kill();
+                                        break;
+                                    }
+                                };
```
rmccorm4, line 163: nit: This would probably benefit from some refactor as it nests pretty deep

rmccorm4 (dismissed):
LGTM - only an optional nit comment - and needs merge conflict resolved

runtime/rust/src/pipeline/network/tcp/client.rs
grahamking: Definitely! It was previously inline in a much bigger function, so this is the first step of that refactor.

rmccorm4 (approved):

PR #135, opened 2025-02-08
chore: Add triton_distributed_rs wheel install to container build

rmccorm4:
Changes:
- Fix `triton_distributed_rs` python binding install when using --mount-workspace by using `maturin build` instead of `maturin develop` - see https://github.com/triton-inference-server/triton_distributed/pull/132#issuecomment-2644718849
    - current code without this PR works with `./container/run.sh -it`, but fails with addition of `--mount-workspace`
    - this PR fixes support for both with or without `--mount-workspace`
- Use release build of crate instead of debug build with `-r` in `maturin build -r`
- Optimize unnecessary docker rebuilds from COPY commands
- Add sanity test to check binding is installed in container build 

Closes #148

PR #157, opened 2025-02-11
feat: Add KV publisher and receiver. Add KV aware routing example.

GuanLuo:
#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID:

- closes GitHub issue: #xxx


container/Dockerfile
```diff
@@ -28,9 +28,38 @@ USER root
...
+RUN cd runtime/rust && \
+    cargo build --release --locked && cargo doc --no-deps
+
+# Generate C bindings
+RUN cd runtime/rust/libtriton_llm && \
+    cargo build --release --locked && cargo doc --no-deps
+
+# Generate Python bindings
+RUN cd runtime/rust/python-wheel && \
+    maturin build
```
rmccorm4, line 53: Merged https://github.com/triton-inference-server/triton_distributed/pull/135 which should contain similar commands to build and install triton_distributed_rs wheel in the container

container/Dockerfile
```diff
@@ -28,9 +28,38 @@ USER root
...
-RUN pip install maturin
+RUN pip install maturin[patchelf]
 ENV PATH="/root/.cargo/bin:${PATH}"
 
+# 'etcd' is runtime dependency
+RUN wget https://github.com/etcd-io/etcd/releases/download/v3.5.18/etcd-v3.5.18-linux-amd64.tar.gz && tar -xzf etcd-v3.5.18-linux-amd64.tar.gz
+RUN cp ./etcd-v3.5.18-linux-amd64/etcd* /usr/local/bin/.
+
+# Working directory
+WORKDIR /workspace
```
rmccorm4, line 39: I also rearranged some of the `COPY` commands in #135 which should reduce the amount of re-builds that happen from unrelated changes.

examples/python/llm/tensorrtllm/README.md
```diff
@@ -243,3 +243,36 @@ Zhang. Distserve: Disaggregating prefill and decoding for goodput-optimized larg
...
+  --request-plane-uri ${HOSTNAME}:4222 &
+```
+
+```bash
+curl ${HOSTNAME}:8000/v1/chat/completions \
+  -H "Content-Type: application/json" \
+  -d '{
+    "model": "llama-3.1-8b-instruct",
+    "messages": [
+      {"role": "user", "content": "Why is Roger Federer the greatest tennis player of all time? Roger Federer is widely regarded as one of the greatest tennis players of all time, and many consider him the greatest."}
```
rmccorm4, line 268: Is this intentionally including the answer to the question too?
```suggestion
      {"role": "user", "content": "Why is Roger Federer the greatest tennis player of all time?"}
```

examples/python/llm/tensorrtllm/operators/kv_aware_routing.py
```diff
@@ -0,0 +1,197 @@
...
+
+import numpy
+
+from triton_distributed.runtime import (
+    RemoteInferenceRequest,
+    RemoteOperator,
+    TritonCoreOperator,
+)
+
+from triton_distributed_rs import KvRouter, DistributedRuntime
```
nnshah1, line 27: nice!

examples/python/llm/tensorrtllm/operators/kv_aware_routing.py
```diff
@@ -0,0 +1,197 @@
...
+        self._preprocess_model = self._triton_core.load("simple_preprocessing")
+        self._postprocess_model = self._triton_core.load("simple_postprocessing")
+        self._logger = logger
+        self._store_outputs_in_response = True
+
+    async def execute(self, requests: list[RemoteInferenceRequest]):
+        self._logger.debug("Executing KvAwareRouting Request")
+        background_tasks = []
+        for request in requests:
+            task = asyncio.create_task(self._execute_request(request))
```
nnshah1, line 59: still not clear to me if this could be simpler - but can look at after - not blocking

@kthui for viz as he is also looking into the worker / operators to see if we can simplify / improve

examples/python/llm/tensorrtllm/operators/kv_aware_routing.py
```diff
@@ -0,0 +1,197 @@
...
+        if "max_tokens" in request.inputs:
+            request_output_len = request.inputs["max_tokens"]
+        elif "max_tokens" in sampling_params:
+            request_output_len = numpy.array(
+                [[sampling_params["max_tokens"]]], dtype=numpy.int32
+            )
+
+        input_ids, input_lengths = await self._preprocess(query)
+        self._logger.debug(input_ids, input_lengths)
+        print(input_ids)
```
nnshah1, line 108:
```suggestion
        self._logger_debug(input_ids)
```
or remove?

examples/python/llm/tensorrtllm/operators/disaggregated_serving.py
```diff
@@ -30,12 +30,12 @@ def __init__(
         self,
         name,
         version,
-        triton_core,
         request_plane,
         data_plane,
         parameters,
         repository,
         logger,
+        triton_core,
```
nnshah1, line 38: thanks!

runtime/rust/libtriton_llm/Cargo.toml
nnshah1, line 1: we have triton-llm for the openai frontend piece as well. possible to combine these into one location?

current non ideal location would be examples/python_rs/llm/ (later to move to examples/llm) - 

@grahamking for viz / suggestion - 

Can wait until after merge but ideally before release tag

nnshah1 (dismissed):
demo changes look good - 

@ryanolson , @grahamking , @rmccorm4  for rust review

ryanolson (changes requested):
llm bits should move to the /llm/rust in the root directory.

runtime should be llm agnostic.

GuanLuo:
> llm bits should move to the /llm/rust in the root directory.
> 
> runtime should be llm agnostic.

@ryanolson So both the `kv_router` and `libtriton_llm` (C binding) should be moved under `llm/rust/triton-llm`? Then I think the `llm/rust/triton-llm` needs to include `python-wheel` similar to `runtime/rust`, for the Python binding of `kv_router`? Is my understanding correct?

ryanolson:
> > llm bits should move to the /llm/rust in the root directory.
> > runtime should be llm agnostic.
> 
> @ryanolson So both the `kv_router` and `libtriton_llm` (C binding) should be moved under `llm/rust/triton-llm`? Then I think the `llm/rust/triton-llm` needs to include `python-wheel` similar to `runtime/rust`, for the Python binding of `kv_router`? Is my understanding correct?

Yes.

kv_router can be a module in triton-llm

libtriton_llm should be a crate in the workspace.

container/Dockerfile
rmccorm4, line 1: needs merge conflicts fixed

python-wheel/rust/llm/kv.rs
```diff
@@ -0,0 +1,52 @@
...
+            Ok(Self { inner })
+        })
+    }
+
+    fn schedule<'p>(
+        &self,
+        py: Python<'p>,
+        token_ids: Vec<u32>,
+        lora_id: u64,
+    ) -> PyResult<Bound<'p, PyAny>> {
```
rmccorm4, line 42: Would it make sense for this to return `PyString` here instead of `PyAny` for a stricter match with the python equivalent definition of `schedule -> str` in `_core.pyi` and stricter compile time checks?

llm/rust/triton-llm/src/kv_router/worker.rs
```diff
@@ -0,0 +1,106 @@
...
+
+use anyhow::Result;
+use derive_builder::Builder;
+use nim_llm_pipeline::network::{
+    ingress::push_endpoint::PushEndpoint,
+    nats::{self, ServiceExt},
+    PushWorkHandler,
+};
+// [FIXME] used new nats?
+// use nova_distributed::transports::nats;
```
rmccorm4, line 28:
```suggestion
// use triton_distributed::transports::nats;
```

llm/rust/triton-llm/src/kv_router/scheduler.rs
```diff
@@ -0,0 +1,287 @@
...
+        log::trace!("awaiting the start of the background endpoint subscriber");
+        let mut endpoints = match endpoints_rx.recv().await {
+            Some(endpoints) => endpoints,
+            None => {
+                return Err(KvSchedulerError::SubscriberShutdown);
+            }
+        };
+
+        // Channel to accept new scheduling requests
+        let (request_tx, request_rx) = tokio::sync::mpsc::channel::<SchedulingRequest>(16);
```
rmccorm4, line 110: @ryanolson started recently tuning some of these channel buffer sizes - in case there's some insight if 16 is OK here or not.

llm/rust/triton-llm/src/kv_router/scheduler.rs
```diff
@@ -0,0 +1,287 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Service {
+    pub name: String,
+    pub id: String,
+    pub version: String,
+    pub started: String,
+    pub endpoints: Vec<Endpoint>,
+}
+
+pub const BLOCK_SIZE: usize = 64;
```
rmccorm4, line 75: I'm guessing this is related to that "min 64 tokens for kv event" requirement from the chat completions example. Can we document it here a little bit? Why 64, and when would you tune it?

llm/rust/triton-llm/src/kv_router.rs
```diff
@@ -0,0 +1,196 @@
...
+mod scoring;
+
+use crate::kv_router::{
+    indexer::{KvIndexer, KvIndexerInterface, RouterEvent},
+    scheduler::{Endpoint, KvScheduler, Service, BLOCK_SIZE},
+    scoring::ProcessedEndpoints,
+};
+
+// this should be discovered from the backend
+pub const KV_BLOCK_SIZE: u64 = 64;
```
rmccorm4, line 36: Is this used anywhere? I don't see it mentioned anywhere else on a `grep`.

And if so, how does it relate to `scheduler.rs::BLOCK_SIZE`?

llm/rust/libtriton-llm/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+
+    let mut token_offset: usize = 0;
+    for block_idx in 0..num_blocks {
+        let block_hash = unsafe { *block_ids.offset(block_idx.try_into().unwrap()) };
+        let tokens = unsafe { token_ids.offset(token_offset.try_into().unwrap()) };
+        let num_toks = unsafe { *num_block_tokens.offset(block_idx.try_into().unwrap()) };
+        // compute hash only apply to full block (64 token) [FIXME] 64 => KV_TOKEN_SIZE
+        if num_toks < 64 {
+            break;
+        }
```
rmccorm4, line 180: Similar question about documenting the min 64 token for kv event requirement

llm/rust/triton-llm/src/kv_router/protocols.rs
```diff
@@ -0,0 +1,205 @@
...
+/// lora_id of a block.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, Ord, PartialOrd)]
+pub struct LocalBlockHash(pub u64);
+
+/// A sequence aware hash of a block where the hash is computed from the tokens_ids, extra_token_ids
+/// and the optional lora_id of a block, PLUS the hash of the parent block.
+///
+/// In this case, the hashing function is external and unknown.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, Ord, PartialOrd)]
+pub struct ExternalSequenceBlockHash(pub u64);
```
rmccorm4, line 36: Just for my own learning - are you using `Copy` instead of `Clone` on some of these structs like LocalBlockHash and ExternalSequenceBlockHash because they're basically just aliases for a primitive type like `u64`?

```rust
let hash1 = LocalBlockHash(12345);

// Copy - implicit, but guaranteed "cheap" implementation
let hash2 = hash1;

// Clone - explicit, but can be user-defined "cheap" or "expensive" implementation
let hash3 = hash1.clone();
```

My current understanding is that `Copy` is guaranteed to be "cheap" (`memcpy` or equivalent for example) and cannot have a custom implementation, whereas `Clone` may be "cheap" or "expensive" depending on how it was implemented. 

My assumption is that the default derived `Clone` would probably be inexpensive (not familiar with the details yet), but a con is that it leaves room for a developer to add a custom implementation themselves and possibly make it expensive later. 

This is just a long way of saying, I'm still trying to figure out the subtleties of when one would Copy or Clone, and curious if you made a concious decision here.

GuanLuo: I think at least for this case, they are the same. And I didn't write this code, CCing author @jthomson04

llm/rust/triton-llm/src/kv_router/scheduler.rs
GuanLuo: 64 is magic number from TRT-LLM, which impacts the implementation. I think it can be relax but haven't looked too much into it. However that needs to be aligned both in the publisher / subscriber which add additional effect if it's made configurable.

GuanLuo:
@ryanolson Do you mind to take a look when you have time?

alec-flowers (dismissed):

llm/rust/triton-llm/src/kv_router/protocols.rs
paulhendricks: @rmccorm4 cc @ryanolson @grahamking 

I think a good rule of thumb with Rust libraries I've interacted with is that by default, they should strive to provide common traits for structs and enums. One of the challenges one encounters is the "orphan rule" - that you can't implement a (foreign) Trait that you don't own for some (foreign) type that you don't own. So if some struct `ForeignStruct`  lives in `foreign_crate`, and `SomeTrait` lives in `another_crate`, you could not do:

```rust
// Does not compile!
// error[E0117]: only traits defined in the current crate can be implemented for types defined outside of the crate
impl another_crate::SomeTrait for foreign_crate::ForeignStruct {
   ...
}
```

What this means in practice is that when developing libraries, it's good practice to provide standard traits like:

1. `Debug`, `Default`, `Clone`, 
2. `Serialize`, `Deserialize`, 
3. `PartialEq`, `Eq`, `PartialOrd`, `Ord`, `Hash `

When possible and where appropriate as users cannot implement those on your types. I've listed them in that order as most users expect them in that order (it's OK if something isn't `Hash` for a good reason but something should almost always have `Debug` or `Clone`). There are ways around this with the Newtype pattern but that's a discussion for another time.

`Copy` is a tricky one as users typically don't expect an object to be `Copy`. If you want a copy of something, you just call `.clone()`. And deriving `Copy` changes the move semantics which may not be obvious.

nnshah1 (approved):

PR #159, opened 2025-02-11
fix: async access to gil updates; notes on perf

ryanolson:
#### What does the PR do?

- improved error handling on deserialize
- offloading gil access on tokio async threads to help avoid starvation
  - this has extra overhead
  - does not fully solve all problems as there are still some rust bits that could add cpu overheads


grahamking (approved):

runtime/rust/python-wheel/rust/engine.rs
```diff
@@ -131,12 +133,18 @@ where
...
-        let stream = Python::with_gil(|py| {
-            let py_request = pythonize(py, &request)?;
-            let gen = self.generator.call1(py, (py_request,))?;
-            let locals = TaskLocals::new(self.event_loop.bind(py).clone());
-            pyo3_async_runtimes::tokio::into_stream_with_locals_v1(locals, gen.into_bound(py))
-        })?;
+        let generator = self.generator.clone();
+        let event_loop = self.event_loop.clone();
+
+        let stream = tokio::task::spawn_blocking(move || {
```
rmccorm4, line 138: Could we add a small comment here for future newbie readers here? For example, (correct me where I'm wrong), is the key point of this block's diff that it is putting the Python GIL acquisition onto a dedicated task that won't block the async runtime - whereas previously is was blocking the runtime?

rmccorm4: I think this is a very subtle but important detail to the overall system performance that could just look like a simple indentation to the right from the previous code.

ryanolson: Yeah.  I added a comment in the top-level readme, but i'll add a note here.

This does add overheads.  Under low GIL contention, this is more expensive, because we have to schedule another a task on a blocking thread pool, then await it.

Under high GIL contentions, we need to offload it or we will strongly block the async threads.

No way to know, so we pay the price of the GIL - this time two fold.

rmccorm4: I meant can we put a comment like that co-located inline with the code here when someone is scanning through how the code works?

grahamking (approved):

runtime/rust/python-wheel/rust/engine.rs
rmccorm4: Not blocking (pun intended), but I think it would be useful

rmccorm4 (approved):

runtime/rust/python-wheel/rust/engine.rs
ryanolson: https://github.com/triton-inference-server/triton_distributed/pull/169/commits/a293ed9876cd36c1728d70d4db025124b6010237

fixed here

ryanolson: merged too quick.  added to the next PR.

Issue #154, opened 2025-02-10
Proposal: Remove `use tracing as log`


grahamking:
The `tracing` crate has a different interface to the `log` crate. Structured logging is easier.

tracing:
```
tracing::info!(%a_value, ?another_but_debug, "message")
```

log (requires `kv` feature):
```
log::info!(a_value:%, another_but_debug:?; "message") 
```

It's subtle but annoying. The semi-colon at the end, and different ways of specifying Display or Debug.

I propose we don't rename `tracing` as `log`, and make all our log statements explicit: `tracing::info!` or `tracing::debug!`.

grahamking:
There is indeed an existing `log` crate and it's the de-factor standard, in the top-20 most downloaded crates. My impression is  that `tracing` is more popular in large projects these days though.

I standardized us on `tracing` here: https://github.com/triton-inference-server/triton_distributed/pull/161

PR #161, opened 2025-02-11
refactor: Use tracing crate

grahamking:
Previously we pretended it was the `log` crate, but they have subtly different interfaces. See https://github.com/triton-inference-server/triton_distributed/issues/154


nv-anants (approved):
LGTM

Issue #158, opened 2025-02-11
performance of disaggregated prefill


Dimensionzw:
I would like to know the performance of the disaggregated prefill implemented in this project. In the vllm project, disaggregated prefill does not support xpyd, that is, multiple p nodes and multiple D nodes. This will lead to poor performance. Is xpyd supported in the implementation of tensorrtllm and vllm in this project? How is the performance? How is the compatibility with other acceleration methods such as tp?

kkranen:
@Dimensionzw please see PR: https://github.com/triton-inference-server/triton_distributed/pull/167

Dimensionzw:
> [@Dimensionzw](https://github.com/Dimensionzw) please see PR: [#167](https://github.com/triton-inference-server/triton_distributed/pull/167)

Received, I will analyze and test it now. Thank you very much

Dimensionzw:
> [@Dimensionzw](https://github.com/Dimensionzw) please see PR: [#167](https://github.com/triton-inference-server/triton_distributed/pull/167)

After installation, I got three packages like this in my pip environment：
```
triton_distributed.icp             0.2.1.dev0+bd70961 /workspace/icp/python
triton-distributed-rs              0.2.0
triton_distributed.runtime         0.0.1.dev0+unknown /workspace/runtime/python
tritonclient                       2.53.0
tritonfrontend                     2.54.0
tritonserver                       0.0.0
```
However, the package triton_distributed is missing, and an error is reported when starting the model service：
```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/workspace/examples/python/llm/vllm/deploy/__main__.py", line 21, in <module>
    from llm.vllm.operators.vllm import (
  File "/workspace/examples/python/llm/vllm/operators/vllm.py", line 24, in <module>
    from triton_distributed.icp.data_plane import DataPlane
ModuleNotFoundError: No module named 'triton_distributed'
```
How should this situation be solved?

Issue #134, opened 2025-02-08
vllm installation outside of venv in standard container can't detect CUDA


rmccorm4:
```bash
# On host
$ ./container/build.sh
$ ./container/run.sh -it

# In container
$ pip install vllm==0.7.2
$ python3
Python 3.12.3 (main, Nov  6 2024, 18:32:19) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import vllm
INFO 02-08 08:00:49 __init__.py:194] No platform detected, vLLM is running on UnspecifiedPlatform
```

But when done using a `uv venv` and `uv pip install vllm==0.7.2`, it works without issue.

May be something simple like setting of LD_LIBRARY_PATH - but needs some investigation.

PR #167, opened 2025-02-12
feat: vLLM v0.7.2 patch supporting XpYd and heterogeneous P/D parallel configs

ptarasiewiczNV:
#### What does the PR do?

Add patch to vLLM v0.7.2 that supports:
- XpYd,
- heterogeneous P/D parallel configs

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [x] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID:

- closes GitHub issue: #xxx


container/Dockerfile
```diff
@@ -161,6 +128,16 @@ RUN mkdir /opt/triton && \
     uv build && \
     uv pip install dist/triton_distributed_rs*cp312*.whl
 
+# Install patched vllm
+ARG VLLM_REF="v0.7.2"
+ARG VLLM_PATCH="vllm_v0.7.2.patch"
+RUN --mount=type=bind,source=./container/deps/,target=/tmp/deps \
+    if [[ "$FRAMEWORK" == "VLLM" ]]; then \
+        source /opt/triton/venv/bin/activate && \
+        bash /tmp/deps/vllm/install.sh --patch /tmp/deps/vllm/${VLLM_PATCH} --ref ${VLLM_REF} --install-cmd "uv pip install --editable" --use-precompiled --installation-dir /opt/vllm ; \
```
rmccorm4, line 137: Out of curiosity - did you try installing vllm in system environment (not in virtual environment)? Curious if you ran into this issue here: https://github.com/triton-inference-server/triton_distributed/issues/134

Also as long as it's only installed in the virtual environment, there will be a different vllm installed from the tritonserver vllm base container in the system environment, and I believe this sanity test will fail unless either (a) the container build is updated to install vllm in system environment, or (b) the gitlab job is updated to activate the virtual environment before running pytest: 
```python
assert vllm.__version__ == "0.7.3.dev0+g0408efc6.d20250212"
```

container/deps/vllm/tests/test_patch_install.py
```diff
@@ -27,14 +27,4 @@
...
-
-@pytest.mark.skipif(vllm is None, reason="Skipping vllm tests, vllm not installed")
-def test_patch_imports():
-    # Verify patched files have no glaring syntax or import issues
-    import vllm.distributed.data_plane as d
-    import vllm.distributed.kv_cache as k
-
-    # Placeholder to avoid unused import errors or removal by linters
-    assert d, k
+    assert vllm.__version__ == "0.7.3.dev0+g0408efc6.d20250212"
```
rmccorm4, line 30: Just double checking - I see v0.7.2 as base install, but v0.7.3 in resultant patched version check here - is that right?

ptarasiewiczNV: Haven't checked that, but it looks like vllm is bumping the patch version when installing from source. There is probably some vllm build flag we can use to fix the version to what we want, e.g. `v0.7.2patchN` would make sense to me. I will check that later (could be in a separate PR)

container/Dockerfile
rmccorm4: > (b) the gitlab job is updated to activate the virtual environment before running pytest:

edit: instead of gitlab-side change

If we lean into the venv route - we can probably update the Dockerfile to activate the virtual environment when entering the container by default for `FRAMEWORK==VLLM`, and remove the extra `source /opt/triton/venv/bin/activate` steps in the `README` for vllm example.

container/deps/vllm/vllm_v0.7.2.patch
rmccorm4, line 1: FYI: https://github.com/triton-inference-server/triton_distributed/issues/158

container/deps/vllm/tests/test_patch_install.py
rmccorm4: Sounds good to me 👍

container/Dockerfile
ptarasiewiczNV: I haven't tried outside virtual env.

Yes, I also thought about just always activating it. Preferably not in `docker run` as not everyone will use our `run.sh` script, but we could just add set `VIRTUAL_ENV` and add bin to `PATH` in the Dockerfile when `FRAMEWORK==VLLM`

nnshah1: agree we should keep it out of the run script and add it to docker if possible - wasn't sure if we have to modify something for that - but if ENV settings are enough - that can work.

What is the general practice for docker / container though -would we expect everything in the system path?

examples/python_rs/llm/vllm/README.md
```diff
@@ -42,17 +42,13 @@ The example is designed to run in a containerized environment using Triton Distr
 
 ```bash
 # Build image
-./container/build.sh
+./container/build.sh --framework VLLM
```
nnshah1, line 45: nice!

container/Dockerfile
ptarasiewiczNV: We could do that with an entrypoint. But maybe it make sense to split into multiple docker files?
```
Dockerfile
Dockerfile.trtllm
Dockerfile.vllm
...
```
?

container/Dockerfile
```diff
@@ -15,7 +15,6 @@
 
 ARG BASE_IMAGE="nvcr.io/nvidia/tritonserver"
 ARG BASE_IMAGE_TAG="25.01-py3"
-ARG VLLM_WHEEL
```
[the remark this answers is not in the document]
ptarasiewiczNV: In fact vllm example does not use tritonserver at all. Even python native path with @nnshah1 improvements. We could even use `vllm/vllm-openai:vX.X.X` as base. Which makes sense in my mind. Let's bring triton dist to the existing inference environment and not the other way around. And let's show that.

container/Dockerfile
ptarasiewiczNV: I have created and issue for that:
https://jirasw.nvidia.com/browse/DLIS-7997
let's make that change in a separate PR

container/Dockerfile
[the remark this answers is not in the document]
ptarasiewiczNV: changed to tritonserver base for the time being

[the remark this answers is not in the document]
rmccorm4: follow-up sounds good for least friction, easier to separate the changes

container/Dockerfile
rmccorm4: follow-up sounds good, but may need to disable the `test_patch_install` until we activate venv by default then

ptarasiewiczNV: It should pass now with the base triton server image, as it will be skipped anyway because no other vllm is installed.

rmccorm4: Ooh that's right, works for me.

examples/python_rs/llm/vllm/common/client.py
```diff
@@ -14,11 +14,11 @@
 # limitations under the License.
 
 
+import argparse
 import asyncio
 
 import uvloop
 from triton_distributed_rs import DistributedRuntime, triton_worker
-from vllm.utils import FlexibleArgumentParser
```
rmccorm4, line 21: Nice removal of vllm req on client side 🚀

rmccorm4 (approved):
LGTM, but we should restore some kind of CI testing for the vllm path / automate running this example in CI ASAP

PR #169, opened 2025-02-12
fix: tcp retry and error handling updates

ryanolson:
#### What does the PR do?

- update tcp client to retry if it gets the error that there are no available ephemeral ports
- improve error handling
- moved rust examples to `/examples/rust`
  - using a workspace for all examples

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID:

- closes GitHub issue: #xxx


runtime/rust/src/pipeline/network/ingress/push_endpoint.rs
```diff
@@ -66,7 +66,15 @@ impl PushEndpoint {
...
                     tracing::trace!(worker_id, "handling new request");
                     let result = ingress.handle_payload(req.message.payload).await;
-                    tracing::trace!(worker_id, "request handled: {:?}", result);
+                    match result {
+                        Ok(_) => {
+                            tracing::trace!(worker_id, "request handled: {:?}", result);
+                        }
+                        Err(e) => {
+                            tracing::warn!("Failed to handle request: {:?}", e);
+                            eprintln!("Failed to handle request: {:?}", e);
```
grahamking, line 75: We don't need both right? Unless it's for Python somehow?

runtime/rust/src/pipeline/network/tcp.rs
```diff
@@ -66,18 +66,18 @@ impl From<TcpStreamConnectionInfo> for ConnectionInfo {
 }
 
 impl TryFrom<ConnectionInfo> for TcpStreamConnectionInfo {
-    type Error = String;
```
grahamking, line 69: Yes!

runtime/rust/src/pipeline/network/tcp/client.rs
```diff
@@ -46,34 +47,45 @@ impl TcpClient {
...
+        loop {
+            match TcpStream::connect(address).await {
+                Ok(socket) => {
+                    socket.set_nodelay(true)?;
+                    return Ok(socket);
+                }
+                Err(e) => {
+                    if e.kind() == std::io::ErrorKind::AddrNotAvailable {
+                        tracing::warn!("retry warning: failed to connect: {:?}", e);
+                        eprintln!("retry warning: failed to connect: {:?}", e);
```
grahamking, line 62: Excessive `eprintln!`

runtime/rust/src/pipeline/network/tcp/client.rs
```diff
@@ -46,34 +47,45 @@ impl TcpClient {
...
         info: ConnectionInfo,
-    ) -> Result<StreamSender, String> {
-        let info = TcpStreamConnectionInfo::try_from(info)?;
+    ) -> Result<StreamSender> {
+        let info =
+            TcpStreamConnectionInfo::try_from(info).context("tcp-stream-connection-info-error")?;
         tracing::trace!("Creating response stream for {:?}", info);
 
         if info.stream_type != StreamType::Response {
-            return Err(format!(
```
grahamking, line 69: `anyhow::bail!` creates an error and immediately returns it.

runtime/rust/src/pipeline/network/tcp/client.rs
```diff
@@ -137,29 +171,21 @@ impl TcpClient {
...
+    let mut framed_reader = framed_reader;
+    let mut alive_tx = alive_tx;
     loop {
         tokio::select! {
             msg = framed_reader.next() => {
                 match msg {
                     Some(Ok(two_part_msg)) => {
                         match two_part_msg.optional_parts() {
                            (Some(bytes), None) => {
-                                let msg: ControlMessage = match serde_json::from_slice(bytes) {
```
grahamking, line 154: No! You lost the `unwrap()` removal.

runtime/rust/src/pipeline/network/tcp/client.rs
```diff
@@ -172,17 +198,22 @@ async fn control_handler(
...
                            _ => {
-                               // we should not receive this
+                                // not a control message, so we just continue
+                               continue;
                            }
                         }
                     }
-                    Some(Err(e)) => {
-                        panic!("failed to decode message from stream: {:?}", e);
+                    Some(Err(_)) => {
```
grahamking, line 206: Can we log the error, just in case?

runtime/rust/src/pipeline/network/ingress/push_endpoint.rs
ryanolson: i was enabling the eprintln for debug builds because not all tests and integrations tests have logging enabled.

triton.code-workspace
```diff
@@ -0,0 +1,22 @@
+{
```
grahamking, line 1: Did you intend to check this in? Seems useful so guessing yes.

runtime/rust/src/pipeline/network/tcp/client.rs
ryanolson: It does. Let's mark this a style point that should come to a consensus on and then address across the code base.

runtime/rust/src/pipeline/network/tcp/client.rs
ryanolson: fixed

triton.code-workspace
ryanolson: yes. it's very helpful. piggy backing this mr

examples/rust/Cargo.toml
```diff
@@ -0,0 +1,54 @@
...
+# limitations under the License.
+
+[workspace]
+members = [
+    "hello_world",
+]
+resolver = "2"
+
+[workspace.package]
+version = "0.1.3"
```
rmccorm4, line 23: To keep in sync with others: https://github.com/triton-inference-server/triton_distributed/pull/155

```suggestion
version = "0.2.0"
```

llm/rust/triton-llm/src/http/service/service_v2.rs
```diff
@@ -60,10 +67,12 @@ impl HttpService {
...
         let observer = cancel_token.child_token();
 
-        Ok(axum::serve(listener, router)
+        axum::serve(listener, router)
             .with_graceful_shutdown(observer.cancelled_owned())
             .await
-            .inspect_err(|_| cancel_token.cancel())?)
+            .inspect_err(|_| cancel_token.cancel())?;
+
+        Ok(())
```
rmccorm4, line 75: These are functionally equivalent, and just a idiomatic thing to put Ok() at the end instead of wrapping around the axum::serve call, right?

runtime/rust/examples/hello_world/rust-toolchain.toml
rmccorm4, line 1: @grahamking I see this deleted but not moved - does it need to move?

runtime/rust/python-wheel/uv.lock
```diff
@@ -114,6 +99,21 @@ wheels = [
     { url = "https://files.pythonhosted.org/packages/63/37/3e32eeb2a451fddaa3898e2163746b0cffbbdbb4740d38372db0490d67f3/pydantic_core-2.27.2-pp310-pypy310_pp73-win_amd64.whl", hash = "sha256:7e17b560be3c98a8e3aa66ce828bdebb9e9ac6ad5466fba92eb74c4c95cb1151", size = 2004715 },
 ]
 
+[[package]]
+name = "triton-distributed-rs"
+version = "0.1.3"
```
rmccorm4, line 104: I think this is already package is already `0.2.0` on main - should it be updated here?

runtime/rust/src/pipeline/network/tcp/client.rs
```diff
@@ -46,34 +47,45 @@ impl TcpClient {
...
-            .await
-            .map_err(|e| format!("failed to connect: {:?}", e))?;
-
-        socket
-            .set_nodelay(true)
-            .map_err(|e| format!("failed to set nodelay: {:?}", e))?;
-
-        Ok(socket)
+    async fn connect(address: &str) -> std::io::Result<TcpStream> {
+        // try to connect to the address; retry with exponential backoff if AddrNotAvailable
```
rmccorm4, line 51: > exponential backoff

1. Where is the error of backoff increasing over time?
2. Is there a max retry count?

runtime/rust/src/pipeline/network/tcp/client.rs
ryanolson: fixed

runtime/rust/tests/soak.rs
```diff
@@ -107,19 +116,42 @@ mod integration {
...
+            let mut tasks = Vec::new();
+            for _ in 0..batch_load {
+                let client = client.clone();
+                tasks.push(tokio::spawn(async move {
+                    let mut stream = tokio::time::timeout(
+                        Duration::from_secs(30),
+                        client.random("hello world".to_string().into()),
+                    )
+                    .await
+                    .context("request timedout")??;
```
rmccorm4, line 132:
```suggestion
                    .context("request timed out")??;
```

runtime/rust/tests/soak.rs
```diff
@@ -107,19 +116,42 @@ mod integration {
...
+
+                    while let Some(_resp) =
+                        tokio::time::timeout(Duration::from_secs(30), stream.next())
+                            .await
+                            .context("stream timedout")?
```
rmccorm4, line 137:
```suggestion
                            .context("stream timed out")?
```

llm/rust/triton-llm/src/http/service/service_v2.rs
ryanolson: Yeah. I felt it was a little cleaner.  If `axum::serve` returned an anyhow::Error, we could just drop the `?;` and `Ok(())` altogether.

runtime/rust/src/transports/etcd.rs
```diff
@@ -99,15 +99,22 @@ impl Client {
...
             etcd_client::Client::connect(config.etcd_url, config.etcd_connect_options).await?;
-        let lease_client = client.lease_client();
 
-        let lease = create_lease(lease_client, 10, token)
-            .await
-            .context("creating primary lease")?;
+        let lease_id = if config.attach_lease {
+            let lease_client = client.lease_client();
+
+            let lease = create_lease(lease_client, 10, token)
```
rmccorm4, line 106: What's the magic number `10` here? Can it be a named var/const?

runtime/rust/examples/hello_world/rust-toolchain.toml
ryanolson: I don't think we need to enforce the tool chain in examples

just in the main sources.  since the examples depend on those, it should inherit, right?  

can we put this in a .cargo at the top level so we don't have to update it in many places?

runtime/rust/src/transports/etcd.rs
```diff
@@ -260,13 +267,18 @@ pub struct ClientOptions {
 
     #[builder(default)]
     etcd_connect_options: Option<ConnectOptions>,
+
+    /// If true, the client will attach a lease to the primary [`CancellationToken`].
```
rmccorm4, line 271: What happens when a lease is not attached (`false`)? Anything the developer needs to know or be aware of?

runtime/rust/python-wheel/uv.lock
ryanolson: uv.lock should be auto generated right?  @alec-flowers ?

runtime/rust/src/transports/etcd.rs
```diff
@@ -99,15 +99,22 @@ impl Client {
...
+        let lease_id = if config.attach_lease {
+            let lease_client = client.lease_client();
+
+            let lease = create_lease(lease_client, 10, token)
+                .await
+                .context("creating primary lease")?;
+
+            lease.id
+        } else {
+            0
```
rmccorm4, line 112: magic number `0` here - should this be `DEFAULT_LEASE_ID` or `INVALID_LEASE_ID` or something?

runtime/rust/src/pipeline/network/tcp/client.rs
ryanolson: i forgot to remove "exponential" i dont think we want it to be exponential.

runtime/rust/src/transports/etcd.rs
ryanolson: let's add that to the etcd config object

runtime/rust/src/pipeline/network/tcp/server.rs
```diff
@@ -374,12 +414,12 @@ async fn tcp_listener(
...
         mut reader: FramedRead<tokio::io::ReadHalf<tokio::net::TcpStream>, TwoPartCodec>,
         writer: FramedWrite<tokio::io::WriteHalf<tokio::net::TcpStream>, TwoPartCodec>,
-    ) -> Result<(), String> {
+    ) -> Result<()> {
         let response_stream = state
             .lock().await
             .rx_subjects
             .remove(&subject)
-            .ok_or(format!("Subject not found: {subject}; upstream publisher specified a subject unknown to the downsteam subscriber"))?;
+            .ok_or(error!("Subject not found: {}; upstream publisher specified a subject unknown to the downsteam subscriber", subject))?;
```
rmccorm4, line 422: I think @grahamking recently made this style change of `"{subject}"` instead of `"{}", subject`, but this looks like it undoes it. Any preference?

runtime/rust/src/transports/etcd.rs
ryanolson: created issue #172

runtime/rust/src/pipeline/network/tcp/server.rs
```diff
@@ -293,20 +298,56 @@ async fn tcp_listener(
...
+    // let listener: TcpListener = socket.into();
+    // let listener = tokio::net::TcpListener::from_std(listener)?;
+
+    // let addr = listener
+    //     .local_addr()
+    //     .map_err(|e| anyhow::anyhow!("Failed get SocketAddr: {:?}", e))?;
+
+    // read_tx
+    //     .send(Ok(addr.port()))
+    //     .expect("Failed to send ready signal");
```
rmccorm4, line 326: needed?

runtime/rust/src/pipeline/network/tcp/server.rs
```diff
@@ -338,19 +380,18 @@ async fn tcp_listener(
         let first_message = framed_reader
             .next()
             .await
-            .ok_or("Connection closed without a ControlMessge".to_string())?
-            .map_err(|e| e.to_string())?;
+            .ok_or(error!("Connection closed without a ControlMessge"))??;
```
rmccorm4, line 383:
```suggestion
            .ok_or(error!("Connection closed without a ControlMessage"))??;
```

runtime/rust/src/transports/etcd.rs
ryanolson: This is primary for cli tools that will want to read etcd, but not be tied to a deployment that require a lease.

I was thinking of using a different decorator or different convenience method to instantiate the defaults for a "worker" and an "observer/tool/etc" ... not really sure what the high level category for these woudl be called.

No need to set up a lease and heartbeat and watch for a cli tool.

runtime/rust/src/transports/etcd.rs
ryanolson: That is a magic number.  0 means no lease in etcd parlance.

ryanolson: We could return this as `None`

runtime/rust/src/pipeline/network/tcp/server.rs
```diff
@@ -140,8 +147,6 @@ impl TcpStreamServer {
                 PipelineError::Generic(format!("Failed to start TcpStreamServer: {}", e))
             })?;
 
-        tracing::info!("TcpStreamServer started on {}:{}", local_ip, local_port);
-
```
rmccorm4, line 144: I think this is a helpful message if it would otherwise start silently. Can it be a debug! message or something if `info` is too loud?

runtime/rust/src/pipeline/network/tcp/client.rs
```diff
@@ -192,24 +232,29 @@ async fn control_handler(
...
         }
     }
     drop(alive_rx);
-    if let Err(e) = framed_writer.get_mut().shutdown().await {
-        tracing::trace!("failed to shutdown writer: {:?}", e);
-    }
+    // if let Err(e) = framed_writer.get_mut().shutdown().await {
+    //     eprintln!("failed to shutdown writer: {:?}", e);
+    //     tracing::trace!("failed to shutdown writer: {:?}", e);
+    // }
```
rmccorm4, line 258: why commented out?

runtime/rust/src/pipeline/network/tcp/server.rs
ryanolson: fixed

runtime/rust/src/pipeline/network/tcp/client.rs
ryanolson: moved that to a watcher task that joins the two and tries to shutdown.

I wonder if this `TIME-WAIT` issue is more a local issue. if the client were on another machine, then perhaps the race on the closure.

it might be that the server should hold the socket open until the client shutdowns, which might avoid the TIME-WAIT state.

runtime/rust/src/pipeline/network/tcp/server.rs
ryanolson: https://github.com/triton-inference-server/triton_distributed/issues/173

ryanolson:
```suggestion
    // TODO(#173) - alternative / not fully functional exploration for #173; removed when resolved.
    // let socket = Socket::new(Domain::IPV4, Type::STREAM, None)?;

    // // Set the socket options
    // socket.set_reuse_address(true)?;
    // socket.set_nodelay(true)?;

    // let addr: SocketAddr = addr.parse()?;
    // //let addr: SocketAddr = "[::1]:0".parse()?;
    // socket.bind(&addr.into())?;

    // socket.listen(128)?;

    // let listener: TcpListener = socket.into();
    // let listener = tokio::net::TcpListener::from_std(listener)?;

    // let addr = listener
    //     .local_addr()
    //     .map_err(|e| anyhow::anyhow!("Failed get SocketAddr: {:?}", e))?;

    // read_tx
    //     .send(Ok(addr.port()))
    //     .expect("Failed to send ready signal");
```

runtime/rust/src/pipeline/network/tcp/client.rs
rmccorm4: Can be follow-up, but I think we'd want a max retry count at some point in an event where connection endlessly fails

runtime/rust/src/pipeline/network/tcp/server.rs
ryanolson: resolving and will follow up later

runtime/rust/src/transports/etcd.rs
rmccorm4: Happy with `None` or `NO_LEASE` or something

runtime/rust/src/pipeline/network/tcp/client.rs
rmccorm4: Where is this fixed? I only see:
```rust
                    Some(Err(_)) => {
                        break;
                    }
```

rmccorm4 (commented):
Generally LGTM. 

Have a few nit comments and would prefer removing the commented out code blocks if not intended to be brought back soon.

Deferring to @grahamking's expertise for approval otherwise

rmccorm4 (approved):

PR #174, opened 2025-02-12
feat: Add `tio` your friendly cmd line uncle to run triton-llm services

grahamking:
This provides a simple example of how to write a triton-llm engine, and how to connect it to the OpenAI HTTP server.

This is the tool previously called `nio` and `llmctl`.

- **Inputs**: Text and HTTP.
- **Engines**: Echo, which streams your prompt back with a slight delay.

Build: `cargo build`

Pre-requisites: `nats-server` and `etcd` must be running locally, even though they are not yet used by `tio`.

Run with text input:
```
./target/debug/tio in=text out=echo_full --model-name test
```

Run with the triton-llm HTTP server:
```
./target/debug/tio in=http out=echo_full --http-port 8080 --model-name Echo-0B
```

List models:
```
curl localhost:8080/v1/models | jq
```

Will output
```
{
  "object": "list",
  "data": [
    {
      "id": "Echo-0B",
      "object": "object",
      "created": 1739400430,
      "owned_by": "nvidia"
    }
  ]
}
```

#### Checklist
- [x] Verified that the PR passes existing CI.
- [x] Verified copyright is correct on all changed files.
- [x] Added _succinct_ git squash message before merging [ref](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
- [x] All template sections are filled out.

It's a binary, so the best place to start is at the top `apps/tio/src/main.rs` in the `main` function. Most of the code (there's not that much) is in `apps/tio/src/lib.rs`.

#### What's next

As triton-distributed gains features `tio` will be able to grow:
- When we get the pre-processor we can have token-in token-out engines. 
- When we get a pull-router we can have `in=nats` and `out=nats`.
- When we get discovery we can have dynamic engines.


apps/tio/src/input/text.rs
```diff
@@ -0,0 +1,151 @@
...
+            tracing::info!("Model: {service_name}");
+            (service_name, engine, false)
+        }
+    };
+    // TODO: Acquire an etcd lease, we are running
+
+    main_loop(cancel_token, &service_name, engine, inspect_template).await
+}
+
+const IS_A_TTY: i32 = 1;
```
ryanolson, line 53: nit: move all consts to the top

apps/tio/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+}
+
+pub async fn run(
+    in_opt: Input,
+    out_opt: Output,
+    flags: Flags,
+    cancel_token: CancellationToken,
+) -> anyhow::Result<()> {
+    // Create the engine matching `out`
+    let engine_config = match out_opt {
```
ryanolson, line 53: neat.  we could add a downstream `{namespace}.{component}.{endpoint}` and just forward the bits

apps/tio/src/output/echo_full.rs
```diff
@@ -0,0 +1,122 @@
...
+
+fn delta_full(id: usize, token: String) -> Annotated<ChatCompletionResponseDelta> {
+    create_delta(id, token, None)
+}
+
+fn stop(id: usize) -> Annotated<ChatCompletionResponseDelta> {
+    create_delta(id, "".to_string(), Some(FinishReason::stop))
+}
+
+fn create_delta(
```
ryanolson, line 88: there is a DeltaGenerator in llm::protocols::openai

ryanolson (changes requested):
use the delta generator from protocols.  if that doesn't work, let's reassess it's usefulness.

apps/tio/src/input/text.rs
grahamking: Done

apps/tio/src/output/echo_full.rs
grahamking: Thanks! Done.

nnshah1 (changes requested):
will need to move llm first into apps - also unclear what tio is / goal

nnshah1 (changes requested):
want to place this in applications/llm/ 

- but need to make a proposal on it and should align with deploy for vllm

apps/llm/tio/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+
+pub async fn run(
+    in_opt: Input,
+    out_opt: Output,
+    flags: Flags,
+    cancel_token: CancellationToken,
+) -> anyhow::Result<()> {
+    // Create the engine matching `out`
+    let engine_config = match out_opt {
+        Output::EchoFull => EngineConfig::StaticFull {
```
ptarasiewiczNV, line 54: @grahamking @ryanolson question: Do I understand correctly that connecting the service to triton workers would simply require adding something like `Output::TritonEndpoint` and passing namespace, component, endpoint through flags? `Output::TritonEndpoint` would then forward the input to the endpoint and stream back the results. Just have to make sure that the protocols will match. Is that correct?

grahamking: Mostly. You would need to implement `AsyncEngine`. Take a look at `output/echo_full.rs` in this PR. That does the "make sure the protocols match" part.

ryanolson (approved):

apps/llm/tio/src/logging.rs
```diff
@@ -0,0 +1,48 @@
...
+static INIT: Once = Once::new();
+
+/// ENV used to set the log level
+const FILTER_ENV: &str = "RUST_LOG";
+
+/// Default log filter, anything RUST_LOG can take
+const DEFAULT_DIRECTIVE: &str = "info";
+
+/// Setup logging. You won't see any output unless you run this.
+pub fn init() {
```
ryanolson, line 31: i just added a `logging::init` to trd.

ryanolson (changes requested):
one change - use the logging from triton_distributed

apps/llm/tio/src/logging.rs
grahamking: Perfect. Using it now.

apps/llm/tio/src/lib.rs
ryanolson: @ptarasiewiczNV - this is what I had in mind

@grahamking - we might want to in a follow up mr have `tio` do `in=http` and `out=dynamic`.

@grahamking - we will also need tio to configure ns/component/ep names -- maybe you have that already.

and maybe `llmctl` does this

```
nio in=http,ns=public-api out=dynamic &

nio --namespace=foo in=pre  out=trtllm &

llmctl http --ns=public-api add-chat-model foo pre.generate
```

llmctl would wait for `foo.pre.generate` to be ready, then write the ModelEntry to the `/trd/ns/public-api/http/models/foo/` which would create a ClientEngine linked to the model named `foo` which would route all requests for that model to any `foo.pre.generate` endpoint(s)

ryanolson (approved):

apps/llm/tio/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+[package]
+name = "tio"
+version = "0.1.0"
+edition = "2021"
+authors = ["NVIDIA"]
+homepage = "https://gitlab-master.nvidia.com/dl/ai-services/libraries/rust/nim-nvllm"
```
rmccorm4, line 21: GitHub link instead?

nnshah1 (approved):
LGTM - thanks for the discussion

PR #177, opened 2025-02-13
fix: Fix rust pre-merge for new tio directory

rmccorm4:
Rust pre-merge check on source code for `tio` moved from `apps/tio` to `applications/llm/bin/tio` - so updating the check accordingly.

Should fix this error: https://github.com/triton-inference-server/triton_distributed/actions/runs/13318456550/job/37198020744

PR #178, opened 2025-02-13
feat: Add a mistralrs engine to tio

grahamking:
This allows us to run a real model.

Build:
```
cargo build --release --features mistralrs,cuda
```

Run:
```
./target/release/tio in=text out=mistralrs --model-path Llama-3.2-1B-Instruct-Q4_K_M.gguf
```

Why [mistral.rs](https://github.com/EricLBuehler/mistral.rs)?

- It has no dependencies. You don't need a container or a virtual env to get started.
- It supports CUDA, Metal (MacOS) and CPU-only. Everyone can join the AI revolution.
- It starts fast and serves fast (with CUDA). That makes it fun to experiment with.
- It runs many models, not just Mistral, that's just it's name.

The key file is `llm/rust/triton-llm/src/engines/mistralrs.rs`. The other changes are small.

### DIFF
--- a/applications/llm/bin/tio/README.md
+++ b/applications/llm/bin/tio/README.md
@@ -50,7 +50,7 @@ List the models: `curl localhost:8080/v1/models`
 
 Send a request:
 ```
-curl -d '{"model": "Llama-3.2-3B-Instruct-Q6_K", "max_tokens": 2048, "messages":[{"role":"user", "content": "What is the capital of South Africa?" }]}' -H 'Content-Type: application/json' http://localhost:8080/v1/chat/completions
+curl -d '{"model": "Llama-3.2-1B-Instruct-Q4_K_M", "max_tokens": 2049, "messages":[{"role":"user", "content": "What is the capital of South Africa?" }]}' -H 'Content-Type: application/json' http://localhost:8080/v1/chat/completions
 ```
 
 Run `tio --help` for more options.
--- a/applications/llm/bin/tio/README.md
+++ b/applications/llm/bin/tio/README.md
@@ -1,0 +1,57 @@
+# triton-llm service runner
+
+`tio` is a tool for exploring the triton-distributed and triton-llm components.
+
+## Install and start pre-requisites
+
+Rust:
+```bash
+curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
+```
+
+Get the NATS server from https://nats.io/download/ and run it:
+```
+nats-server -js --trace --store_dir $(mktemp -d)
+```
+
+Get etcd from https://github.com/etcd-io/etcd/releases and run it: `etcd`
+
+These components are required but not yet used by tio. It's a journey, OK.
+
+## Build
+
+- CUDA:
+
+`cargo build --release --features mistralrs,cuda`
+
+- MAC w/ Metal:
+
+`cargo build --release --features mistralrs,metal`
+
+- CPU only:
+
+`cargo build --release --features mistralrs`
+
+## Download a model from Hugging Face
+
+For example one of these should be fast and good quality on almost any machine: https://huggingface.co/bartowski/Llama-3.2-3B-Instruct-GGUF
+
+## Run
+
+*Text interface*
+
+`./target/release/tio in=text out=mistralrs --model-path Llama-3.2-1B-Instruct-Q4_K_M.gguf`
+
+*HTTP interface*
+
+`./target/release/tio in=http out=mistralrs --model-path Llama-3.2-1B-Instruct-Q4_K_M.gguf`
+
+List the models: `curl localhost:8080/v1/models`
+
+Send a request:
+```
+curl -d '{"model": "Llama-3.2-3B-Instruct-Q6_K", "max_tokens": 2048, "messages":[{"role":"user", "content": "What is the capital of South Africa?" }]}' -H 'Content-Type: application/json' http://localhost:8080/v1/chat/completions
+```
+
+Run `tio --help` for more options.
+
--- a/applications/llm/bin/tio/src/main.rs
+++ b/applications/llm/bin/tio/src/main.rs
@@ -20,10 +20,15 @@ use clap::Parser;
 use triton_distributed::logging;
 
 const HELP: &str = r#"
-triton-llm service runner stub
+triton-llm service runner
+
+Example:
+- cargo build --release --features mistralrs,cuda
+- ./target/release/tio in=text out=mistralrs --model-path Llama-3.2-1B-Instruct-Q4_K_M.gguf --model-name 'Llama-3.2-1B-Instruct'
+
 "#;
 
-const USAGE: &str = "USAGE: tio in=[http|text] out=[echo_full] [--http-port 8080]";
+const USAGE: &str = "USAGE: tio in=[http|text] out=[mistralrs|echo_full] [--http-port 8080] [--model-path <path>] [--model-name <served-model-name>]";
 
 fn main() -> anyhow::Result<()> {
     logging::init();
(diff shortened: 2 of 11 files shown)


grahamking:
That CODEOWNERS is pretty intense! Hello reviewers, I didn't tag you, that's github, but I'm glad you're here.

rmccorm4:
FYI the rust pre-merge check isn't currently running on the `tio` code - added a fix here: https://github.com/triton-inference-server/triton_distributed/pull/177

applications/llm/bin/tio/src/main.rs
```diff
@@ -20,10 +20,15 @@ use clap::Parser;
 use triton_distributed::logging;
 
 const HELP: &str = r#"
-triton-llm service runner stub
+triton-llm service runner
+
+Example:
+- cargo build --release --features mistralrs,cuda
```
rmccorm4, line 26: CC @nv-anants @nvda-mesharma I think we need a way to kick off rust builds & tests that require GPU on gitlab side

grahamking:
> FYI the rust pre-merge check isn't currently running on the `tio` code - added a fix here: #177

Thank you. Rebased.

nnshah1:
@harryskim, @dmitry-tokarev-nv  for viz - will want to make sure attributions files updated

applications/llm/bin/tio/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+*HTTP interface*
+
+`./target/release/tio in=http out=mistralrs --model-path Llama-3.2-1B-Instruct-Q4_K_M.gguf`
+
+List the models: `curl localhost:8080/v1/models`
+
+Send a request:
+```
+curl -d '{"model": "Llama-3.2-3B-Instruct-Q6_K", "max_tokens": 2048, "messages":[{"role":"user", "content": "What is the capital of South Africa?" }]}' -H 'Content-Type: application/json' http://localhost:8080/v1/chat/completions
```
paulhendricks, line 53: Just ran the example: `{"error":"Model not found"}⏎`

Can we change this to `Llama-3.2-3B-Instruct-Q4_K_M`?

applications/llm/bin/tio/src/main.rs
```diff
@@ -20,10 +20,15 @@ use clap::Parser;
...
+- ./target/release/tio in=text out=mistralrs --model-path Llama-3.2-1B-Instruct-Q4_K_M.gguf --model-name 'Llama-3.2-1B-Instruct'
+
 "#;
 
-const USAGE: &str = "USAGE: tio in=[http|text] out=[echo_full] [--http-port 8080]";
+const USAGE: &str = "USAGE: tio in=[http|text] out=[mistralrs|echo_full] [--http-port 8080] [--model-path <path>] [--model-name <served-model-name>]";
```
paulhendricks, line 31: Perhaps change `<path>` to `<full-path>` or `<absolute-path>` - I tried with `~/workspace/Llama...` and tripped for a minute thinking my GGUF file was corrupted.

Or perhaps add helpful context to error message about using full path.

llm/rust/triton-llm/src/engines/mistralrs.rs
```diff
@@ -0,0 +1,278 @@
...
+            return_raw_logits: false,
+        });
+
+        self.mistralrs.get_sender()?.send(mistralrs_request).await?;
+
+        let mut used_output_tokens = 0;
+        let output = stream! {
+            loop {
+                let response = match rx.recv().await {
+                    Some(res) => res.as_result().unwrap(),
```
paulhendricks, line 215: Do we care about `unwrap()`s?

@rmccorm4 - Do we want to consider adding a clippy linting rule e.g. `clippy::unwrap`?

paulhendricks (commented):
Great work Graham!

llm/rust/triton-llm/src/engines/mistralrs.rs
```diff
@@ -0,0 +1,278 @@
...
+            tool_choice: None,
+            logits_processors: None,
+            return_raw_logits: false,
+        });
+
+        self.mistralrs.get_sender()?.send(mistralrs_request).await?;
+
+        let mut used_output_tokens = 0;
+        let output = stream! {
+            loop {
```
paulhendricks, line 213: Maybe consider `while let` instead of `loop`? Reads a bit cleaner and lets us handle `Err` over unwrap?

```rust
let output = stream! {
    while let Some(response) = rx.recv().await {
        match response.as_result() {
            Ok(ResponseOk::Chunk(c)) => { ... }
            Err(e) => {
                tracing::error!("Error: {:?}", e);
                break;
            }
            x => tracing::error!("Unhandled. {x:?}"),
        }
    }
};
```

llm/rust/triton-llm/src/engines/mistralrs.rs
nnshah1: for my edification - is unwrap bad?

llm/rust/triton-llm/src/engines/mistralrs.rs
nnshah1: @paulhendricks is there a style guide / static format / check we can reference ?

llm/rust/triton-llm/src/engines/mistralrs.rs
rmccorm4: unwrap() will [panic!](https://doc.rust-lang.org/std/macro.panic.html) on error (abort/exit)

When something returns a Value or Error (Result/Option), calling unwrap() on it is like calling `assert not error`

rmccorm4: > @rmccorm4 - Do we want to consider adding a clippy linting rule e.g. clippy::unwrap?

@paulhendricks it sounds like a nice thing to automate to me to avoid shooting ourselves in the foot with runtime panics where we don't want them (non-fatal errors) and isn't unnecessarily restrictive or something. 

Would that be this one specifically? https://rust-lang.github.io/rust-clippy/master/index.html#unwrap_used

@grahamking @ryanolson thoughts? Any other clippy checks we should add as well?

paulhendricks: @nnshah1 - `unwrap()` lets you develop quickly without needing to explicitly handle every single error which can be a bit tedious. Though RM mentioned, you typically want to handle errors gracefully instead of unwrapping an error and your entire program panicking and shutting down.

And @rmccorm4 - yes. Another good one is `deny(missing_docs)` but again the tradeoff there is development speed vs. consistent documentation.

It's also possible to add clippy as a pre-commit step: https://doc.rust-lang.org/clippy/usage.html

applications/llm/bin/tio/README.md
grahamking: Fixed, thank you.

applications/llm/bin/tio/src/main.rs
grahamking: This is great feedback. I made it take either full or relative paths.

llm/rust/triton-llm/src/engines/mistralrs.rs
grahamking: *static checks*

So far I've had to replace `don't` with `do not` in a comment because the spell check couldn't handle apostrophes, and I've had to exclude `README.md` from needing a copyright notice.

Let's enforce clean clippy using the default settings, I think we already do that. No need to add extra rules just yet.

*unwrap*

Good catch. I will handle all of those, or add a "Safety" comment when the unwrap is clean.

llm/rust/triton-llm/src/engines/mistralrs.rs
grahamking: Very nice! Fixed.

paulhendricks (approved):

PR #184, opened 2025-02-14
feat: Add Dockerfile.vllm

ptarasiewiczNV:
#### What does the PR do?

Add separate `Dockerfile.vllm` for vLLM example. The uv virtual env is activated in the Dockerfile.vllm.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID:

- closes GitHub issue: #xxx

### DIFF
--- a/container/Dockerfile
+++ b/container/Dockerfile
@@ -128,16 +128,6 @@ RUN mkdir /opt/triton && \
     uv build && \
     uv pip install dist/triton_distributed_rs*cp312*.whl
 
-# Install patched vllm
-ARG VLLM_REF="v0.7.2"
-ARG VLLM_PATCH="vllm_v0.7.2.patch"
-RUN --mount=type=bind,source=./container/deps/,target=/tmp/deps \
-    if [[ "$FRAMEWORK" == "VLLM" ]]; then \
-        source /opt/triton/venv/bin/activate && \
-        bash /tmp/deps/vllm/install.sh --patch /tmp/deps/vllm/${VLLM_PATCH} --ref ${VLLM_REF} --install-cmd "uv pip install --editable" --use-precompiled --installation-dir /opt/vllm ; \
-    fi
-
-
 # Install triton_distributed_rs wheel globally in container for tests that
 # currently run without virtual environment activated.
 # TODO: In future, we may use a virtualenv for everything and remove this.
--- a/container/Dockerfile.vllm
+++ b/container/Dockerfile.vllm
@@ -1,0 +1,119 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+ARG BASE_IMAGE="nvcr.io/nvidia/tritonserver"
+ARG BASE_IMAGE_TAG="25.01-py3"
+
+FROM ${BASE_IMAGE}:${BASE_IMAGE_TAG} AS triton-distributed
+
+USER root
+
+# Rust build/dev dependencies
+RUN apt-get update; apt-get install -y gdb protobuf-compiler
+RUN curl https://sh.rustup.rs -sSf | bash -s -- -y
+ENV PATH="/root/.cargo/bin:${PATH}"
+
+# Install OpenAI-compatible frontend
+# TODO: can this be removed since we have rust http server?
+ARG OPENAI_SERVER_TAG="r25.01"
+RUN mkdir -p /opt/tritonserver/python && \
+    cd /opt/tritonserver/python && \
+    rm -rf openai && \
+    git clone -b ${OPENAI_SERVER_TAG} --single-branch https://github.com/triton-inference-server/server.git && \
+    cd server && \
+    git checkout ${SERVER_OPENAI_COMMIT} && \
+    cd .. && \
+    mv server/python/openai openai && \
+    chown -R root:root openai && \
+    chmod 755 openai && \
+    chmod -R go-w openai && \
+    rm -rf server && \
+    python3 -m pip install -r openai/requirements.txt
+
+# Common dependencies
+RUN --mount=type=bind,source=./container/deps/requirements.txt,target=/tmp/requirements.txt \
+    pip install --timeout=2000 --requirement /tmp/requirements.txt
+RUN --mount=type=bind,source=./container/deps/requirements.nats.txt,target=/tmp/requirements.txt \
+    pip install --timeout=2000 --requirement /tmp/requirements.txt
+RUN --mount=type=bind,source=./container/deps/requirements.test.txt,target=/tmp/requirements.txt \
+    pip install --timeout=2000 --requirement /tmp/requirements.txt
+
+# Finish pyright install
+RUN pyright --help > /dev/null 2>&1
+
+# In Process Python API Install
+# TODO: can this be removed since we do not use the tritonserver?
+RUN find /opt/tritonserver/python -maxdepth 1 -type f -name \
+    "tritonserver-*.whl" | xargs -I {} pip3 install --force-reinstall --upgrade {}[all]
+
+# GENAI Perf Install
+ARG GENAI_PERF_TAG="r25.01"
+RUN pip install "git+https://github.com/triton-inference-server/perf_analyzer.git@${GENAI_PERF_TAG}#subdirectory=genai-perf"
+
+# Install NATS
+RUN wget https://github.com/nats-io/nats-server/releases/download/v2.10.24/nats-server-v2.10.24-amd64.deb && dpkg -i nats-server-v2.10.24-amd64.deb
+
+# Enable Git operations in the /workspace directory
+RUN printf "[safe]\n      directory=/workspace\n" > /root/.gitconfig
+
+RUN ln -sf /bin/bash /bin/sh
+
+# Install NGINX and utilities
+# TODO: can this be removed since we do not use the nginx?
+RUN apt-get install nginx nvtop tmux -y
+RUN rm -rf /etc/nginx/sites-enabled/default
+
+# Working directory
+WORKDIR /workspace
+
+COPY runtime /workspace/runtime
+RUN cd runtime/rust && \
+    cargo build --release --locked && cargo doc --no-deps
+
+COPY llm/rust /workspace/llm/rust
+COPY examples/rust /workspace/examples/rust
+RUN cd examples/rust && \
+    cargo build --release && \
+    cp target/release/http /usr/local/bin/ && \
+    cp target/release/llmctl /usr/local/bin/
+
+# Install uv and create virtualenv
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
+RUN mkdir /opt/triton && \
+    uv venv /opt/triton/venv --python 3.12 && \
+    source /opt/triton/venv/bin/activate && \
+    cd runtime/rust/python-wheel && \
+    uv build && \
+    uv pip install dist/triton_distributed_rs*cp312*.whl
+
+# Install patched vllm
+ARG VLLM_REF="v0.7.2"
+ARG VLLM_PATCH="vllm_v0.7.2.patch"
+RUN --mount=type=bind,source=./container/deps/,target=/tmp/deps \
+    source /opt/triton/venv/bin/activate && \
+    bash /tmp/deps/vllm/install.sh --patch /tmp/deps/vllm/${VLLM_PATCH} --ref ${VLLM_REF} --install-cmd "uv pip install --editable" --use-precompiled --installation-dir /opt/vllm
+
+# Install triton_distributed_rs wheel globally
+RUN pip install runtime/rust/python-wheel/dist/triton_distributed_rs*cp312*.whl
+
+COPY icp /workspace/icp
+RUN /workspace/icp/protos/gen_python.sh
+
+# Install python packages
+ARG PYTHON_PACKAGE_VERSION=0.0.1.dev+unknown
+RUN SETUPTOOLS_SCM_PRETEND_VERSION_FOR_TRITON_DISTRIBUTED_ICP=${PYTHON_PACKAGE_VERSION} pip install -e /workspace/icp/python
+RUN SETUPTOOLS_SCM_PRETEND_VERSION_FOR_TRITON_DISTRIBUTED_RUNTIME=${PYTHON_PACKAGE_VERSION} pip install -e /workspace/runtime/python
+
+# Copy remaining files
+COPY . /workspace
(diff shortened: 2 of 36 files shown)


ptarasiewiczNV:
Some parts of the Dockerfile.vllm are probably fine to remove, but I wasn't sure what is used by native python example so just put TODOs for now.

container/Dockerfile.vllm
```diff
@@ -0,0 +1,112 @@
...
+    git clone -b ${OPENAI_SERVER_TAG} --single-branch https://github.com/triton-inference-server/server.git && \
+    cd server && \
+    git checkout ${SERVER_OPENAI_COMMIT} && \
+    cd .. && \
+    mv server/python/openai openai && \
+    chown -R root:root openai && \
+    chmod 755 openai && \
+    chmod -R go-w openai && \
+    rm -rf server && \
+    python3 -m pip install -r openai/requirements.txt
```
rmccorm4, line 31: If we use the rust http server instead of the pure python.llm.api_server, then yes this shouldn't be needed anymore. @nnshah1 for what we want to do the the pure python example for now.

container/Dockerfile.vllm
```diff
@@ -0,0 +1,112 @@
...
+
+# Enable Git operations in the /workspace directory
+RUN printf "[safe]\n      directory=/workspace\n" > /root/.gitconfig
+
+RUN ln -sf /bin/bash /bin/sh
+
+# Install NGINX and utilities
+# TODO: can this be removed since we do not use the nginx?
+RUN apt-get install nginx nvtop tmux -y
+RUN rm -rf /etc/nginx/sites-enabled/default
```
rmccorm4, line 64: @nnshah1

rmccorm4 (commented):
@ptarasiewiczNV can you add `Dockerfile.vllm` to this path list to trigger vllm job: https://github.com/triton-inference-server/triton_distributed/blob/4f07691d82cafa765240bace77f44c09ac90f68b/.github/workflows/trigger_ci.yml#L55-L58 when it gets modified?

(note current pipeline on latest commit excluded vllm jobs)

container/Dockerfile.vllm
```diff
@@ -0,0 +1,112 @@
...
+
+# Install NATS
+RUN wget https://github.com/nats-io/nats-server/releases/download/v2.10.24/nats-server-v2.10.24-amd64.deb && dpkg -i nats-server-v2.10.24-amd64.deb
+
+# Enable Git operations in the /workspace directory
+RUN printf "[safe]\n      directory=/workspace\n" > /root/.gitconfig
+
+RUN ln -sf /bin/bash /bin/sh
+
+# Install NGINX and utilities
```
whoisj, line 61: For what reason, and I honestly cannot think of any, would we want NGINX in the container? Seems like bloatware.

container/Dockerfile.vllm
```diff
@@ -0,0 +1,112 @@
...
+# TODO: can this be removed since we do not use the tritonserver?
+RUN find /opt/tritonserver/python -maxdepth 1 -type f -name \
+    "tritonserver-*.whl" | xargs -I {} pip3 install --force-reinstall --upgrade {}[all]
+
+# GENAI Perf Install
+ARG GENAI_PERF_TAG="r25.01"
+RUN pip install "git+https://github.com/triton-inference-server/perf_analyzer.git@${GENAI_PERF_TAG}#subdirectory=genai-perf"
+
+# Install NATS
+RUN wget https://github.com/nats-io/nats-server/releases/download/v2.10.24/nats-server-v2.10.24-amd64.deb && dpkg -i nats-server-v2.10.24-amd64.deb
```
whoisj, line 54: Why is the nats's .deb being downloaded into the container. I do not see it being used anywhere, in which case it only serves to bloat the size of the already too large container image.

container/Dockerfile.vllm
```diff
@@ -0,0 +1,112 @@
...
+RUN --mount=type=bind,source=./container/deps/requirements.test.txt,target=/tmp/requirements.txt \
+    pip install --timeout=2000 --requirement /tmp/requirements.txt
+
+# Finish pyright install
+RUN pyright --help > /dev/null 2>&1
+
+# In Process Python API Install
+# TODO: can this be removed since we do not use the tritonserver?
+RUN find /opt/tritonserver/python -maxdepth 1 -type f -name \
+    "tritonserver-*.whl" | xargs -I {} pip3 install --force-reinstall --upgrade {}[all]
```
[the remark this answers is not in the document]
whoisj: how much container image size is retained by keeping Triton Server around?

container/Dockerfile.vllm
```diff
@@ -0,0 +1,112 @@
...
+RUN mkdir /opt/triton && \
+    uv venv /opt/triton/venv --python 3.12 && \
+    source /opt/triton/venv/bin/activate && \
+    cd runtime/rust/python-wheel && \
+    uv build && \
+    uv pip install dist/triton_distributed_rs*cp312*.whl
+
+# Install patched vllm
+ARG VLLM_REF="v0.7.2"
+ARG VLLM_PATCH="vllm_v0.7.2.patch"
```
whoisj, line 84: this value should likely be formatted more like
```
ARG VLLM_PATCH="vllm_${VLLM_REF}.patch"
```
as I assume we need both values to use the same version in most cases.

whoisj (commented):
Generally really like this PR, left a few notes/questions. Thanks!

container/Dockerfile.vllm
```diff
@@ -0,0 +1,112 @@
...
+# Copy remaining files
+COPY . /workspace
+
+# Environment setup
+ENV PYTHONPATH="${PYTHONPATH}:/workspace/examples/python:/opt/tritonserver/python/openai/openai_frontend"
+ENV RAPIDS_LIBUCX_PREFER_SYSTEM_LIBRARY=true
+
+# Activate virtual environment
+ENV VIRTUAL_ENV=/opt/triton/venv
+ENV PATH="${VIRTUAL_ENV}/bin:${PATH}"
```
rmccorm4, line 109: Nice add!

For python_rs vllm example I think it will work as-is right now, but the container ends up in a bit of an unintuitive state regarding python and pip I think:

Setup/run container
```
$ ./container/build.sh --framework vllm
$ ./container/run.sh --framework vllm -it
```

Explore python/pip setup
```
# unintuitive: No (venv) in shell PS1 prompt, probably due to order of operations with bashrc/profile
# but we are in fact inside the virtual environment
root@ced35d0-lcedt:/workspace# which python3
/opt/triton/venv/bin/python3

# OK - vllm installed in virtual environment as expected
root@ced35d0-lcedt:/workspace# python3 -c "import vllm"
INFO 02-14 20:51:17 __init__.py:190] Automatically detected platform cuda.
WARNING 02-14 20:51:17 cuda.py:336] Detected different devices in the system: 
WARNING 02-14 20:51:17 cuda.py:336] NVIDIA GeForce RTX 3090
WARNING 02-14 20:51:17 cuda.py:336] NVIDIA RTX 5880 Ada Generation
WARNING 02-14 20:51:17 cuda.py:336] Please make sure to set `CUDA_DEVICE_ORDER=PCI_BUS_ID` to avoid unexpected behavior.

# unintuitive: `pip` is not the virtualenv's pip
root@ced35d0-lcedt:/workspace# which pip
/usr/bin/pip

# virtualenv has no `pip`
root@ced35d0-lcedt:/workspace# python3 -m pip
/opt/triton/venv/bin/python3: No module named pip

# virtualenv does have `uv pip`
root@ced35d0-lcedt:/workspace# uv pip freeze | grep vllm
Using Python 3.12.3 environment at: /opt/triton/venv
-e file:///opt/vllm

# No pure python libs installed in the virtual environment (no python icp/runtime access)
# (same as /opt/triton/venv/bin/python3)
root@ced35d0-lcedt:/workspace# python3 -c "import triton_distributed"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'triton_distributed'

# But they are installed in system environment
root@ced35d0-lcedt:/workspace# /usr/bin/python3 -c "import triton_distributed" # OK

# Python bindings to rust are installed in both environments as expected
root@ced35d0-lcedt:/workspace# /usr/bin/python3 -c "import triton_distributed_rs"             # OK
root@ced35d0-lcedt:/workspace# /opt/triton/venv/bin/python3 -c "import triton_distributed_rs" # OK
```

TLDR; 
- Great: There is a set of things installed in the venv, and the venv is activated by default when entering the container - this works great
- Confusing: `pip` is the system pip, not the virtual environment pip (`uv pip` works in venv for those familiar)
    - for example `pip freeze` does not represent the set of things installed in the environment, and mentions packages like triton_distributed.icp that aren't installed in the venv, and _doesn't_ mention packages like `vllm` which _are_ installed in the venv
- Confusing: If trying to run the pure python workflows:
    - with the lack of `(venv)` in the shell prompt, I'm not sure the user knows that they are in a virtual environment or not
    - there doesn't seem to be a `deactivate` available on the PATH to get out of the virtual environment if needed

So I think we should iron out some things like:
- Make `pip` reference the virtualenv pip (or `uv pip`?) when inside the virtualenv
- Decide how to handle the two sets of dependencies - all in virtual environment? two virtual environments? steps to deactivate virtual environment?

container/Dockerfile.vllm
nnshah1: lets leave it for now - we can remove later once we have the end to end set up - we can also compare functionality.

If we need to tweak something it might be easier in the python verison -

container/Dockerfile.vllm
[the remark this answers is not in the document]
nnshah1: agree with @rmccorm4 - as long as its in the container - lets keep this line. If we change bases - we will not add it back.

container/Dockerfile.vllm
nnshah1: lets do a seperate minimization step - right now this is a mix of runtime and dev dependencies - we'll need to refactor more carefully

container/Dockerfile.vllm
nnshah1: see above- originally we use this for dev simplicitly of running our baseline nginix examples. 

Agree we need to be more thoughtful movign forward. - but I think seperate from this PR

container/Dockerfile.vllm
nnshah1: yes can be removed - let's do it seperately

container/Dockerfile.vllm
nnshah1: @rmccorm4 - pre merge or post merge?

rmccorm4: Can be post merge since python_rs example works as-is and pure python example has a NOTE at top of readme that it's not currently tested. I believe both before and after this PR, the pure python example won't run as-is on main based on vLLM patch install changes and such.

nnshah1: can you add a story to track?

container/Dockerfile.vllm
nnshah1: @whoisj - before after merge?

nnshah1 (dismissed):
LGTM 

-if other developers need changes - please mention

@rmccorm4 , @whoisj

nnshah1:
> @ptarasiewiczNV can you add `Dockerfile.vllm` to this path list to trigger vllm job:
> 
> https://github.com/triton-inference-server/triton_distributed/blob/4f07691d82cafa765240bace77f44c09ac90f68b/.github/workflows/trigger_ci.yml#L55-L58
> 
> when it gets modified?
> (note current pipeline on latest commit excluded vllm jobs)

this should be fixed pre-merge - right, @rmccorm4

container/Dockerfile.vllm
ptarasiewiczNV: @whoisj agree

ptarasiewiczNV:
@nnshah1 @rmccorm4 @whoisj I have updated the `Dockerfile.vllm` according to your suggestions. For the tests to pass we would have to install test deps in the virtual env anyway so I simply moved all deps inside the venv (it is activated in the Dockerfile)

nv-anants (approved):

container/Dockerfile
```diff
@@ -128,16 +128,6 @@ RUN mkdir /opt/triton && \
...
 
-# Install patched vllm
-ARG VLLM_REF="v0.7.2"
-ARG VLLM_PATCH="vllm_v0.7.2.patch"
-RUN --mount=type=bind,source=./container/deps/,target=/tmp/deps \
-    if [[ "$FRAMEWORK" == "VLLM" ]]; then \
-        source /opt/triton/venv/bin/activate && \
-        bash /tmp/deps/vllm/install.sh --patch /tmp/deps/vllm/${VLLM_PATCH} --ref ${VLLM_REF} --install-cmd "uv pip install --editable" --use-precompiled --installation-dir /opt/vllm ; \
-    fi
-
```
rmccorm4, line 139: FYI @alec-flowers @sshchoi et al for viz on change to Dockerfile for vllm workflow

container/Dockerfile.vllm
dmitry-tokarev-nv: I second on separating devel image from runtime.
Cuda started doing the same with a 15% drop on the image size:
```
gitlab-master.nvidia.com:5005/dl/dgx/cuda           12.8-devel-ubuntu24.04--25.01      5c74a4dcc35d   3 weeks ago      9.7GB
gitlab-master.nvidia.com:5005/dl/dgx/cuda           12.8-runtime-ubuntu24.04--25.01    570f5974a906   3 weeks ago      8.27GB
```
Triton server will start using cuda runtime images for monthly releases.

container/Dockerfile.vllm
dmitry-tokarev-nv: I'd argue to remove nginx and let users install it as part of running examples.

container/Dockerfile.vllm
```diff
@@ -0,0 +1,111 @@
...
+
+# Install OpenAI-compatible frontend
+# TODO: can this be removed since we have rust http server?
+ARG OPENAI_SERVER_TAG="r25.01"
+RUN mkdir -p /opt/tritonserver/python && \
+    cd /opt/tritonserver/python && \
+    rm -rf openai && \
+    git clone -b ${OPENAI_SERVER_TAG} --single-branch https://github.com/triton-inference-server/server.git && \
+    cd server && \
+    git checkout ${SERVER_OPENAI_COMMIT} && \
```
dmitry-tokarev-nv, line 33: Where do we provide this value?
Why complicate it like this we can just create a tag and checkout a tag.

Issue #139, opened 2025-02-09
Proposal: Remove secondary runtime


grahamking:
It takes control of the Tokio runtime away from the user. I can make a primary runtime with `Runtime::from_handle`, but the secondary runtime creation is hard coded in `Runtime::new`. I don't think runtime creation belongs in a library like this.

It will make debugging harder because it is something almost no-one else does. For example to use tokio-console this would need two console-subscriber.

It don't think it helps. Blocking either runtime is going to be fatal. We'd prevent that with an integration test on a single threaded runtime (default `#[tokio::test]`).

Happy to submit a PR if you're OK with the proposed change.

grahamking:
Moved to PR #185

PR #185, opened 2025-02-14
refactor: Remove secondary runtime and lease

grahamking:
Why remove it?

- It makes the system easier to understand. There's a single tokio runtime which the user can (and should) provide, and there's a single etcd lease.

- It gives control of the Tokio runtime back to the user. I can make a primary runtime with Runtime::from_handle, but the secondary runtime creation is hard coded in Runtime::new. I don't think runtime creation belongs in a library like this.

- It makes debugging easier. Two runtimes is something almost no-one else does. For example to use tokio-console this would need two console-subscriber.

- The second one doesn't help. Blocking either runtime is going to be fatal. We will prevent that with an integration test on a single threaded runtime (default #[tokio::test]).

Closes https://github.com/triton-inference-server/triton_distributed/issues/139


ryanolson (changes requested):
My preference is to hold off on this. 

I want the ability to isolate background/low-priority tasks vs. application tasks and keep them in a separate scheduler.

PR #186, opened 2025-02-14
fix: Unique IDs for mistralrs requests

grahamking:
Upgrade mistralrs to latest.


grahamking:
Thanks to some great out of band feedback from @paulhendricks I was able to improve ID generation.

paulhendricks (approved):
@grahamking and discussed using `next_request_id()` from mistralrs. Looks good 👍

PR #190, opened 2025-02-14
refactor: Proposal to refactor protocols using async-openai

paulhendricks:
#### What does the PR do?

Proposal: Refactor `protocols` using `async-openai` to maintain compatibility with OpenAI and offload maintenance to purpose-built ecosystem crates while still maintaining NVIDIA-specific extensions.

Starting with creating an `NvidiaFinishReason` that extends `async_openai::types::FinishReason` with `Cancelled` and `Null`.

I wanted to start a discussion around using OpenAI protocols and chose using `FinishReason` as a way to demonstrate this. Consider the below.

Before:
```rust
#[derive(Clone, Copy, Debug, Deserialize, Serialize, PartialEq, Eq)]
#[allow(non_camel_case_types)]
pub enum FinishReason {
    stop,
    length,
    content_filter,
    tool_calls,
    cancelled,
    null,
}
```

After:
```rust
use async_openai::types::FinishReason as OpenAIFinishReason;

#[derive(Debug, Clone, Copy, PartialEq)]
pub enum NvidiaFinishReason {
    OpenAI(OpenAIFinishReason),
    Cancelled,
    Null,
}
```

There are pros and cons to this approach. 

Pros:
Offloading maintenance of these bits to purpose-built crates in the Rust. E.g. if OpenAI's API ever changes (which it does quite often), hopefully it is easy for `triton_distributed` - with the right abstractions - to upgrade and maintain OpenAI-compliance while still allowing us the freedom to provide NVIDIA-specific extensions. We can also start small with pieces like `FinishReason` and incrementally migrate over time.

Cons:
As this PR hopefully shows, it is bit more difficult then just swapping out `FinishReason`. In fact, the simple change above changes several other files. And we would want to fully test these changes and their behavior before migrating.

#### Checklist
- [X] PR title reflects the change and is of format `<commit_type>: <Title>`
- [X] Changes are described in the pull request.

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [X] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:
N/A

`./llm/rust/triton-llm/src/protocols/openai/chat_completions.rs`

#### Test plan:
 Not yet but will possibly do after discussion.
 Not yet but will possibly do after discussion.
 N/A

- CI Pipeline ID:

- closes GitHub issue: #xxx


grahamking:
- Were you not able to derive the Serialize / De-serialize for `NvidiaFinishReason`?
- Nit, but `format!("{var}")` instead of `format!("{}", var)`.

PR #194, opened 2025-02-16
[feat, refactor] Weighted random endpoint selection; refactor request sending

jthomson04:
Refactor identical code in the Rust client's request handling, and add support for sending requests to a random endpoint with weighted probabilities.


lib/runtime/src/component/client.rs
```diff
@@ -216,10 +213,48 @@ where
             endpoints[offset as usize]
         };
 
-        let subject = self.endpoint.subject_to(endpoint_id);
-        let request = request.map(|req| AddressedRequest::new(req, subject));
+        self.send_request(endpoint_id, request).await
+    }
 
-        self.router.generate(request).await
+    /// Issue a request to a random endpoint with a weighted selection.
```
ryanolson, line 219: Neat. Are we thinking we use the kv load to capacity ratio here?

PR #201, opened 2025-02-18
feat: event plane + count

ryanolson:
#### What does the PR do?

Count is coded.  Untested.
- High touch MR
- See IPC + EventPlane trait + Namespace w/ EventPlane impl

- Follow on MRs will merge latest KV Router merge with Count - also high touch
- Follow on MRs will apply the EventPlane trait to:
  - DistributedRuntime, Component and Endpoint objects
  - Provide Python bindings for all object updated with the EventPlane Trait.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [x] feat
- [ ] fix
- [ ] perf
- [x] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID:

- closes GitHub issue: #xxx


applications/llm/count/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+#derive_builder = "0.20"
+#futures = "0.3"
+#serde_json = { version = "1" }
+#thiserror = { version = "2.0.11" }
+#tokio = { version = "1", features = ["full"] }
+#tokio-stream = { version = "0.1" }
+#tokio-util = { version = "0.7", features = ["codec", "net"] }
+#tracing = { version = "0.1" }
+#validator = { version = "0.20.0", features = ["derive"] }
+#uuid = { version = "1", features = ["v4", "serde"] }
```
rmccorm4, line 33: needed?
```suggestion
```

applications/llm/count/src/main.rs
```diff
@@ -0,0 +1,142 @@
...
+
+fn get_config() -> Result<LLMWorkerLoadCapacityConfig> {
+    let component_name = std::env::var("NOVA_COUNT_SCRAPE_COMPONENT")?;
+    if component_name.is_empty() {
+        return Err(error!("NOVA_COUNT_SCRAPE_COMPONENT is not set"));
+    }
+
+    let endpoint_name = std::env::var("NOVA_COUNT_SCRAPE_ENDPOINT")?;
+    if endpoint_name.is_empty() {
+        return Err(error!("NOVA_COUNT_SCRAPE_ENDPOINT is not set"));
```
rmccorm4, line 48: I'm sure all the prefixes will be subject to change - but should they be prefixed `TRD_*` to be consistent with some of the other ones you had going? 

Also, any ongoing documentation on all the env vars the system supports anywhere?

applications/llm/count/src/main.rs
```diff
@@ -0,0 +1,142 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Count is a metrics aggregator designed to operate within a namespace and collect
+//! metrics from all workers.
+//!
+//! Metrics will collect for now:
+//!
+//! - LLM Worker Load:Capacity
+//!   - These metrics will be scraped by the LLM NATS Service API's stats request
```
rmccorm4, line 22: Does "LLM NATS Service API" ~= "Event Plane"? Or different?

applications/llm/count/src/main.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    let service_name = target.service_name();
+    let subject = target_endpoint.subject();
+
+    log::debug!("Scraping service {service_name} and filtering on subject {subject}");
+    let token = drt.primary_lease().child_token();
+
+    loop {
+        let next = Instant::now() + Duration::from_secs(2);
+        let stream = target.scrape_stats(Duration::from_secs(1)).await?;
```
rmccorm4, line 116: Should these be based on a configurable polling interval?

applications/llm/count/src/main.rs
```diff
@@ -0,0 +1,142 @@
...
+            .filter(|e| e.subject == subject)
+            .filter_map(|e| match e.data {
+                Some(metrics) => metrics.decode::<LLMWorkerLoadCapacity>().ok(),
+                None => None,
+            })
+            .collect::<Vec<_>>();
+
+        // compute mean / std of LLMWorkerLoadCapacity
+
+        // publish using the namespace event plane
```
rmccorm4, line 128:
```suggestion
        // todo - compute mean / std of LLMWorkerLoadCapacity

        // todo - publish using the namespace event plane
```

examples/rust/service_metrics/README.md
```diff
@@ -0,0 +1,9 @@
+# Service Metrics
+
+This example extends the hello_world example by calling the `scrape_service` method
+with the service name for the request response the client just issued a request.
+
+The client can now observe some basic statistics about each instance of the service
+begin hosted.
+
+If you start two copies of the server, you will see two entries being emitted.
```
rmccorm4, line 9: todo: Show example output here

examples/rust/service_metrics/src/bin/client.rs
```diff
@@ -0,0 +1,67 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use futures::StreamExt;
+use service_metrics::DEFAULT_NAMESPACE;
+
+// use tokio::time::{Duration, Instant};
```
rmccorm4, line 19:
```suggestion
```

examples/rust/service_metrics/src/bin/client.rs
```diff
@@ -0,0 +1,67 @@
...
+    let mut stream = client.random("hello world".to_string().into()).await?;
+
+    while let Some(resp) = stream.next().await {
+        println!("{:?}", resp);
+    }
+
+    let metrics = distributed
+        .nats_client()
+        .scrape_service(&component.service_name())
+        .await?;
```
rmccorm4, line 55: Calling scrape_service on the nats_client seems a little awkward to me for user-facing code. Can this be simplified to something like:
```suggestion
    let metrics = distributed
        .scrape_service(&component.service_name())
        .await?;
```

and internally on the DistributedRuntime method it hides/wraps the nats_client part?
```rust
self.nats_client().scrape_service(...)
```

applications/llm/count/Cargo.toml
ryanolson: I'm just holding those there while developing so it easy to copy them in from tmp. 

But yes, before we merge let's delete the #tmp section

examples/rust/service_metrics/src/bin/server.rs
```diff
@@ -0,0 +1,81 @@
...
+        Ok(ResponseStream::new(Box::pin(stream), ctx.context()))
+    }
+}
+
+async fn backend(runtime: DistributedRuntime) -> Result<()> {
+    // attach an ingress to an engine
+    let ingress = Ingress::for_engine(RequestHandler::new())?;
+
+    // // make the ingress discoverable via a component service
+    // // we must first create a service, then we can attach one more more endpoints
```
rmccorm4, line 69:
```suggestion
    // make the ingress discoverable via a component service
    // we must first create a service, then we can attach one more more endpoints
```

runtime/rust/src/icp/events.rs
```diff
@@ -0,0 +1,43 @@
...
+use std::fmt::Debug;
+
+use crate::Result;
+
+// #[async_trait]
+// pub trait Publisher: Debug + Clone + Send + Sync {
+//     async fn publish(&self, event: &(impl Serialize + Send + Sync)) -> Result<()>;
+// }
+
+/// A [EventPlane] is a component that can publish and/or subscribe to events on the event plane.
```
rmccorm4, line 12: Should this say "on the event plane" or "on <something else>"? Seems weird to refer to itself - "An event plane publishes to events on the event plane"
```suggestion
/// A [EventPlane] is a component that can publish and/or subscribe to events.
```

runtime/rust/src/icp/events.rs
```diff
@@ -0,0 +1,43 @@
...
+        bytes: Vec<u8>,
+    ) -> Result<()>;
+
+    // /// Create a new publisher for the given event name. The `event_name` will be `.` concatenated with the
+    // /// base subject provided by the implementation.
+    // fn publisher(&self, event_name: impl AsRef<str>) -> impl Publisher;
+
+    // /// Create a new publisher for the given event name. The `event_name` will be `.` concatenated with the
+    // fn publisher(&self, event_name: impl AsRef<str>) -> Result<Publisher>;
+    // fn publisher_bytes(&self, event_name: impl AsRef<str>) -> &PublisherBytes;
```
rmccorm4, line 42: When would you need to create a `publisher` vs. just using `event_plane.publish()`?

runtime/rust/src/service.rs
```diff
@@ -85,33 +85,32 @@ impl ServiceClient {
...
-        if !err.is_empty() {
-            tracing::error!("failed to collect services: {:?}", err);
+    pub async fn collect_services(
+        &self,
+        service_name: &str,
+        duration: Duration,
+    ) -> Result<ServiceSet> {
+        let sub = self.nats_client.scrape_service(service_name).await?;
+        if duration.is_zero() || duration > Duration::from_secs(10) {
+            tracing::warn!("collect_services: duration is zero or greater than 10 seconds");
```
rmccorm4, line 95: nit: Make these separate warnings instead of having to reverse engineer whether it was 0 or > 10: `if duration.is_zero() - warn("duration is zero")`, `if duration > 10: warn("duration is greater than 10 seconds")`

runtime/rust/src/service.rs
```diff
@@ -85,33 +85,32 @@ impl ServiceClient {
...
-
-        if !err.is_empty() {
-            tracing::error!("failed to collect services: {:?}", err);
+    pub async fn collect_services(
+        &self,
+        service_name: &str,
+        duration: Duration,
+    ) -> Result<ServiceSet> {
+        let sub = self.nats_client.scrape_service(service_name).await?;
+        if duration.is_zero() || duration > Duration::from_secs(10) {
```
rmccorm4, line 94: question: What is the significance of 10 seconds here? Why is that the cutoff? Can it be given a a definition or motivation with in-line comments here?

runtime/rust/src/transports/nats.rs
```diff
@@ -86,7 +86,14 @@ impl Client {
         Ok(stream)
     }
 
-    pub async fn service_subscriber(&self, service_name: &str) -> Result<Subscriber> {
+    /// Issues a broadcast request for all services with the provided `service_name` to report their
+    /// current stats. Each service will only respond once. The service may have customized the reply
+    /// so the caller should select which endpoint and what concretely data model should be used to
```
rmccorm4, line 91:
```suggestion
    /// so the caller should select which endpoint and what concrete data model should be used to
```

runtime/rust/src/transports/nats.rs
```diff
@@ -86,7 +86,14 @@ impl Client {
...
+    /// extract the details.
+    ///
+    /// Note: Because each endpoint will only reply once, the caller must drop the subscription after
+    /// some time or it will await forever.s
```
rmccorm4, line 95:
```suggestion
    /// some time or it will await forever.
```

examples/rust/service_metrics/src/bin/client.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    let deadline = Instant::now() + Duration::from_millis(200);
+    let mut metrics = stream::until_deadline(metrics, deadline);
```
rmccorm4, line 58: Is this setting a deadline because of the note in the code about dropping the subcription? Or just to avoid waiting indefinitely?

```rust
    /// Note: Because each endpoint will only reply once, the caller must drop the subscription after
    /// some time or it will await forever
```

runtime/rust/src/transports/nats.rs
rmccorm4, line 95: >    /// ... Because each endpoint will only reply once, the caller must drop the subscription after
>    /// some time or it will await forever

This seems like a very important note that can be prone to error. Is there a way for us to enforce this automatically for the user/caller with some sensible default/timeout somewhere instead?

examples/rust/service_metrics/src/bin/client.rs
rmccorm4: This also looks a little similar to `component.scrape_stats()` which calls `drt.sevice_client().collect_services()` which calls `self.nats_client().scrape_service()` - maybe we can simplify some things here

runtime/rust/src/icp/events.rs
```diff
@@ -0,0 +1,43 @@
...
+// #[async_trait]
+// pub trait Publisher: Debug + Clone + Send + Sync {
+//     async fn publish(&self, event: &(impl Serialize + Send + Sync)) -> Result<()>;
+// }
+
+/// A [EventPlane] is a component that can publish and/or subscribe to events on the event plane.
+///
+/// Each implementation of [EventPlane] will define the root subject.
+#[async_trait]
+pub trait EventPlane {
```
rmccorm4, line 16: Where does subscribing to the event plane fit in here? Which pieces should be refactored to subscribe through event plane instead of what they're currently doing? 

I see some spots that are doing `self.nats_client().subscribe()` :
- Are these spots to be refactored?
- Are there others?

applications/llm/count/src/main.rs
ryanolson: yes, eventually.  need a primary configuration object for this app.

examples/rust/service_metrics/src/bin/client.rs
ryanolson: yes, we have to formalize this across the stack. lots to do here.

runtime/rust/src/icp/events.rs
ryanolson: publisher just captures the event name.  in theory, you could create a publisher with a strongly type message, so that both the event name and event type are captured.

it's probably not needed, but could be nice.

runtime/rust/src/icp/events.rs
ryanolson: Correct.  We might need to capture these as issues and resolve them independently.

applications/llm/count/src/main.rs
ryanolson: done

applications/llm/count/src/main.rs
ryanolson: keeping for now.  we will refresh the count main.rs after the refactor and full pipeline integration.

examples/rust/service_metrics/README.md
ryanolson: done

examples/rust/service_metrics/src/bin/client.rs
ryanolson: done

runtime/rust/src/icp/events.rs
ryanolson: save for later

runtime/rust/src/icp/events.rs
```diff
@@ -0,0 +1,58 @@
...
+//     async fn publish(&self, event: &(impl Serialize + Send + Sync)) -> Result<()>;
+// }
+
+/// A [EventPlane] is a component that can publish and/or subscribe to events.
+///
+/// Each implementation of [EventPlane] will define the root subject.
+#[async_trait]
+pub trait EventPlane {
+    /// The base subject used for this implementation of the [EventPlane].
+    fn subject(&self) -> String;
```
nnshah1, line 33:
```suggestion
    fn topic(&self) -> String;
```

runtime/rust/src/transports/nats.rs
ryanolson: It's an internal call, so ultimately, it won't be publicly exposed.

runtime/rust/src/icp/events.rs
ryanolson: save for later.

runtime/rust/src/icp/events.rs
nnshah1, line 1: can we rename to event_plane.rs?

runtime/rust/src/icp/events.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    /// Publish a single event to the event plane. The `event_name` will be `.` concatenated with the
+    /// base subject provided by the implementation.
+    async fn publish(
+        &self,
+        event_name: impl AsRef<str> + Send + Sync,
```
nnshah1, line 39: we had event type - would that be the same as event name here?

we could use event_type / event_topic

runtime/rust/src/icp/events.rs
```diff
@@ -0,0 +1,58 @@
...
+        event: &(impl Serialize + Send + Sync),
+    ) -> Result<()>;
+
+    /// Publish a single event as bytes to the event plane. The `event_name` will be `.` concatenated with the
+    /// base subject provided by the implementation.
+    async fn publish_bytes(
```
nnshah1, line 45: does this need to be a seperate method - or could be part of publish itself?

publish(Event) 

Event.payload is bytes?

runtime/rust/src/icp/events.rs
```diff
@@ -0,0 +1,58 @@
...
+        &self,
+        event_name: impl AsRef<str> + Send + Sync,
```
nnshah1, line 47: I *think* this is event_topic - is that right from the design?

nnshah1 (commented):
Would it be possible to seperate out the event plane interface from the rest of the changes here to make review a little simpler?

runtime/rust/src/icp/events.rs
ryanolson: publish takes any object that is serializable and publishes it as a json encoded bytes.

publish_bytes is any bytes, byo-serialization format.

runtime/rust/src/icp/events.rs
ryanolson: the subject/topic is computed from the `{self.subject()}.events.{event_name}`

runtime/rust/src/icp/events.rs
ryanolson: the trait is called EventPlane. 

it's nice to keep the modules to simpler names.

runtime/rust/src/icp/events.rs
ryanolson: rust doesn't have function overloads like c++

runtime/rust/src/utils/stream.rs
```diff
@@ -0,0 +1,50 @@
...
+};
+
+use tokio::time::{self, sleep_until, Duration, Instant, Sleep};
+
+pub struct DeadlineStream<S> {
+    stream: S,
+    sleep: Pin<Box<Sleep>>,
+}
+
+impl<S: Stream + Unpin> Stream for DeadlineStream<S> {
```
paulhendricks, line 30: Do we need to support `!Unpin` streams? 

I've got some code I'll send you that allows `!Unpin` streams (e.g. `tokio::time::interval` is `!Unpin`) but not sure what the deeper ramifactions of `Unpin` vs. `!Unpin` are here. More of a curiousity than a blocker.

paulhendricks (approved):

nnshah1 (approved):

applications/llm/count/src/main.rs
```diff
@@ -0,0 +1,206 @@
...
+//! - LLM Worker Load:Capacity
+//!   - These metrics will be scraped by the LLM NATS Service API's stats request
+//!   - Request Slots: [Active, Total]
+//!   - KV Cache Blocks: [Active, Total]
+
+use serde::{Deserialize, Serialize};
+
+use triton_distributed::{
+    error,
+    icp::events::EventPlane,
```
nnshah1, line 30: icp::event_plane::EventPlane

applications/llm/count/src/main.rs
```diff
@@ -0,0 +1,206 @@
...
+
+        let processed = ProcessedEndpoints {
+            capacity_with_ids,
+            load_avg,
+            load_std,
+            address: address.clone(),
+        };
+
+        // publish using the namespace event plane
+        namespace.publish(&event_name, &processed).await?;
```
nnshah1, line 178: can we just have this be part of the runtime?

drt.publish_event()

applications/llm/count/src/main.rs
```diff
@@ -0,0 +1,206 @@
...
+                continue;
+            }
+        }
+    }
+
+    Ok(())
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ProcessedEndpoints {
```
nnshah1, line 194: does this end up being a json over the event plane? or msgspec?

examples/rust/service_metrics/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
nnshah1, line 1: service metrics is a library ?

examples/rust/service_metrics/README.md
```diff
@@ -0,0 +1,25 @@
...
+Annotated { data: Some("l"), id: None, event: None, comment: None }
+Annotated { data: Some("l"), id: None, event: None, comment: None }
+Annotated { data: Some("o"), id: None, event: None, comment: None }
+Annotated { data: Some(" "), id: None, event: None, comment: None }
+Annotated { data: Some("w"), id: None, event: None, comment: None }
+Annotated { data: Some("o"), id: None, event: None, comment: None }
+Annotated { data: Some("r"), id: None, event: None, comment: None }
+Annotated { data: Some("l"), id: None, event: None, comment: None }
+Annotated { data: Some("d"), id: None, event: None, comment: None }
+ServiceSet { services: [ServiceInfo { name: "triton_init_backend_720278f8", id: "j6n37goJog3df2PMkQK1Ry", version: "0.0.1", started: "2025-02-18T20:51:01.40830026Z", endpoints: [EndpointInfo { name: "triton_init_backend_720278f8-generate-694d94fc30dbb562", subject: "triton_init_backend_720278f8.generate-694d94fc30dbb562", data: Some(Metrics(Object {"average_processing_time": Number(67387), "last_error": String(""), "num_errors": Number(0), "num_requests": Number(1), "processing_time": Number(67387), "queue_group": String("q")})) }] }] }
```
nnshah1, line 24: what is triton_init_backend here?

llm/rust/triton-llm/src/kv_router/scheduler.rs
```diff
@@ -24,6 +24,7 @@ use crate::kv_router::indexer::OverlapScores;
 pub use crate::kv_router::protocols::{ForwardPassMetrics, KV_BLOCK_SIZE};
 use crate::kv_router::scoring::ProcessedEndpoints;
 
+#[allow(dead_code)]
```
nnshah1, line 27: maybe a comment here to say why we want to allow dead code?

runtime/rust/src/icp/events.rs
nnshah1: but that doesn't match the icp.event_plane naming scheme -

if its easy to do let's rename

nnshah1 (changes requested):
I'd like to split this up into smaller PRs and have one that focuses just on event plane and make it match:

https://github.com/triton-inference-server/triton_distributed/blob/main/icp/python/src/triton_distributed/icp/event_plane.py

dmitry-tokarev-nv (approved):
No concerns from my side. The PR predominantly has Rust code changes. Deferring Rust code review to core developers.

llm/rust/triton-llm/src/kv_router/scheduler.rs
ryanolson: I put that in there to avoid all the warnings.  This was a quick fix which should have been caught by CI.  CI should enforce the build to complete without errors. This happened on the KV Router merge.  We should open a ticket and resolve those in a separate PR.

examples/rust/service_metrics/README.md
ryanolson: the name generated from the namespace and component.

examples/rust/service_metrics/Cargo.toml
ryanolson: is an example of how to scrape metrics from components serving endpoints/functions.

applications/llm/count/src/main.rs
ryanolson: json

applications/llm/count/src/main.rs
ryanolson: The core ideas is that the events are published and subscribed to at some level.  

Yes, I expect there to be a `drt.publish`, but also a `my_component.publish`

The difference is the root component of the topic.

We need to make sure the topics to no collide across different levels in the hierarchy.

example:

`drt.publish(event_name, event)` might emit events with the subject `__global__.events.{event_name}`

`ns.publish(event_name, event)` might emit events with the subject `__{namespace}__`.events.{event_name}`

etc.

applications/llm/count/src/main.rs
```diff
@@ -0,0 +1,206 @@
...
+    // todo move to distributed and standardize and move into file/env/cli config
+    let namespace = std::env::var("TRD_NAMESPACE").unwrap_or("default".to_string());
+
+    let namespace = drt.namespace(namespace)?;
+    let component = namespace.component("count")?;
+
+    // there should only be one count
+    // check {component.etcd_path()}/instance for existing instances
+    let key = format!("{}/instance", component.etcd_path());
+    drt.etcd_client()
```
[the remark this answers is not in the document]
ryanolson: The plan is to remove all instances of etcd_client and move those to a DiscoveryPlane facasde/trait, which will behave similarly to EventPlane below.

For now, we need the flexibility to just call etcd, but part of the API readiness is to encapsulate how we end up need to use the client directly.

Personally, I think we should continue to provide access to ectd raw using the 0.x phase and deprecate it in a year or so as the API solidifies and we get closer to a 1.0

runtime/rust/src/utils/stream.rs
ryanolson: Definitely something to explore and determine best practices on.

.github/workflows/pre-merge-rust.yml
```diff
@@ -28,7 +28,7 @@ on:
     paths:
     - 'runtime/rust/**'
     - 'llm/rust/**'
-    - 'applications/llm/bin/tio/**'
+    - 'applications/llm/tio/**'
```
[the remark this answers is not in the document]
ryanolson: i'm not sure why there is a bin directory anyway.  All the applications are binaries or in the case of python entrypoints.

Given the fact we are reorganizing anyway, it's probably moot.

applications/llm/count/src/main.rs
ryanolson: there is more to events than just the trait.

runtime/rust/src/icp/events.rs
ryanolson: it's also redundant. i prefer simplicity.

runtime/rust/src/icp/events.rs
ryanolson: I will add, that there are probably two levels of subscription.  One in which you get the payload and the second which you get the subject and any other bits from the nats message we might want to forward on.

Imagine you are matching on a wild card, you might want to parse the subject to get the values that matched the glob pattern and to a combined (key, value) action.

nnshah1 (changes requested):
as discussed offline -let's remove icp from directory - and keep events to a traits / interface folder in runtime - 

we'll focus on two libraries - llm and runtime and defocus on icp - to simplify the api surface -

piotrm-nvidia (changes requested):
Pls adjust GitHub action to trigger gitlab tests. It is not tested.

piotrm-nvidia (dismissed):

applications/llm/count/src/main.rs
```diff
@@ -0,0 +1,206 @@
...
+
+use triton_distributed::{
+    error,
+    icp::events::EventPlane,
+    logging,
+    utils::{Duration, Instant},
+    DistributedRuntime, ErrorContext, Result, Runtime, Worker,
+};
+
+use tracing as log;
```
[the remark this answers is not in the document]
ishandhanani: 😂😂😂

runtime/rust/src/component/namespace.rs
```diff
@@ -0,0 +1,70 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    // todo - make a distributed runtime fixture
+    // todo - two options - fully mocked or integration test
+    #[cfg(feature = "integration")]
+    #[tokio::test]
+    async fn test_publish() {
```
rmccorm4, line 60: FYI @tmonty12 another integration test being added here - should get caught by the same feature flag you were looking at adding to CI

rmccorm4 (approved):
May need to generally add a couple more `cargo test`'s to the pre-merge-rust (llm/rust, mabye here, etc) - but @tmonty12 was looking into that, so LGTM

nv-anants (approved):
github workflow changed look good

grahamking (approved):

PR #203, opened 2025-02-18
bug: tio can run without NATS or ETCD

grahamking:
Previously `tio` wouldn't run unless etcd and nats were present in the environment, even though it doesn't use them yet.

This makes it easier to use `tio`.

### DIFF
--- a/applications/llm/bin/tio/src/main.rs
+++ b/applications/llm/bin/tio/src/main.rs
@@ -77,15 +77,9 @@ async fn tio_wrapper(runtime: triton_distributed::Runtime) -> anyhow::Result<()>
 
     // etcd and nats addresses, from env vars ETCD_ENDPOINTS and NATS_SERVER with localhost
     // defaults
-    let dt_config = triton_distributed::distributed::DistributedConfig::from_settings();
+    //let dt_config = triton_distributed::distributed::DistributedConfig::from_settings();
     // Wraps the Runtime (which wraps two tokio runtimes) and adds etcd and nats clients
-    let d_runtime = triton_distributed::DistributedRuntime::new(runtime, dt_config).await?;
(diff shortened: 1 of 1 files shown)


ryanolson (approved):

rmccorm4 (approved):

applications/llm/bin/tio/src/main.rs
```diff
@@ -77,15 +77,9 @@ async fn tio_wrapper(runtime: triton_distributed::Runtime) -> anyhow::Result<()>
 
     // etcd and nats addresses, from env vars ETCD_ENDPOINTS and NATS_SERVER with localhost
     // defaults
-    let dt_config = triton_distributed::distributed::DistributedConfig::from_settings();
+    //let dt_config = triton_distributed::distributed::DistributedConfig::from_settings();
     // Wraps the Runtime (which wraps two tokio runtimes) and adds etcd and nats clients
-    let d_runtime = triton_distributed::DistributedRuntime::new(runtime, dt_config).await?;
+    //let d_runtime = triton_distributed::DistributedRuntime::new(runtime, dt_config).await?;
```
rmccorm4, line 82: Can you elaborate a bit on the implications of this? For example, with the distributed_runtime included, would you be able to hook up components exposed run by `tio` with `llmctl` or other tooling, and with these commented out, would they no longer be discoverable? Looking for high level ideas to learn more about how the moving parts relate to each other

grahamking: There are no implications. We were creating a DistributedRuntime but not using any parts of it. 

Currently there are no components and no discovery. Once we start doing `out=nats` we will want the distributed runtime, so I left it in commented out.

ryanolson (approved):

applications/llm/bin/tio/src/main.rs
rmccorm4: Makes sense, thanks Graham!

PR #208, opened 2025-02-18
build: fix wheel naming and add icp dependencies

nv-anants:
* match namespace packages naming convention (use `-` instead of `.` in name)
https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages
```
[build-system]
...

[tool.setuptools.packages.find]
where = ["src/"]
include = ["mynamespace.subpackage_a"]

[project]
name = "mynamespace-subpackage-a"
```

* Add minimal dependencies for icp, required to do a successful import after whl install
ToDo: remove redundant deps (decalared as icp dependency) from container/deps files 

* Add license. ToDo: figure out a way to use top level license file for all builds. maturin fails with relative path
```
1.694 💥 maturin failed
1.694   Caused by: Failed to build source distribution
1.694   Caused by: Failed to add file from /workspace/python-wheel/../LICENSE to sdist as triton_distributed_rs-0.2.1/../LICENSE
1.694   Caused by: paths in archives must not have `..` when setting path for triton_distributed_rs-0.2.1
```


runtime/python/pyproject.toml
```diff
@@ -18,7 +18,7 @@ requires = ["setuptools>=65.0", "setuptools-scm>=8"]
 build-backend = "setuptools.build_meta"
 
 [project]
-name = "triton_distributed.runtime"
+name = "triton_distributed-runtime"
```
ryanolson, line 21: we need some consistency here...

nv-anants: the thought for going with the `triton_distributed-runtime` was to differentiate between the top level import (`triton_distributed`) and subpackage (`runtime`). Distribution name is normalized for `-` and `_`, however common practice is to use `-` i believe. Should we do `triton-distributed-runtime`?

Note - Import statements wont be impacted by this. It stays `import triton_distributed.runtime`

nnshah1: let's use '-' for package names

'_' for file names

@ryanolson - work?

nv-anants: updated to use `-`

nnshah1 (approved):
LGTM

PR #209, opened 2025-02-19
feat: naive heuristic disagg router

tedzhouhk:
#### What does the PR do?
Implement a naive heuristic disagg router in rust and expose to python binding.
The router simply decides to remote locally/remotely based on the length of tokens to prefill.
Marking as draft since conditional disagg is not ready in vllm.

Example usage:
```
>>> from triton_distributed_rs import DisaggregatedRouter
>>> router = DisaggregatedRouter(512)
>>> router.prefill_remote(1000, 200)
True
>>> router.prefill_remote(100, 200)
False
>>> router.update_value(1024)
>>> router.prefill_remote(1000, 200)
False
>>> router.prefill_remote(1500, 200)
True
```

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [x] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID:

- closes GitHub issue: #xxx


examples/python_rs/llm/vllm_nixl/README.md
```diff
@@ -17,6 +17,22 @@ limitations under the License.
 
 > **NOTE**: This example is based on an internal NVIDIA library that will soon be publicly released. The example won't work until the official release.
 
+## Prerequisites
+
+Start required services (etcd and NATS):
+
+   Option A: Using [Docker Compose](/runtime/rust/docker-compose.yml) (Recommended)
```
nnshah1, line 24:
```suggestion
   Option A: Using [Docker Compose](/deploy/docker-compose.yml) (Recommended)
```

examples/python_rs/llm/vllm_nixl/README.md
```diff
@@ -35,42 +51,83 @@ All of the commands below are run inside the same container.
...
+
 ```
-llmctl http add chat-models deepseek-ai/DeepSeek-R1-Distill-Llama-8B test-nixl.vllm.generate
+llmctl http add chat-models deepseek-ai/DeepSeek-R1-Distill-Llama-8B triton-init.process.chat/completions
 TRT_LOG=DEBUG http --port 8181
 ```
 
+To launch disaggregated vllm deployment, there are three major components:
+1. Processor
+2. KV Router
```
nnshah1, line 62: is the kv router required - or optional?

examples/python_rs/llm/vllm_nixl/README.md
```diff
@@ -35,42 +51,83 @@ All of the commands below are run inside the same container.
...
+
+
+### Disaggregated Router
 
-### Disaggregated deployment
+The disaggregated router determines whether a request should be send to a
+remote prefill engine or a local prefill engine for prefilling based on the
+prefill length. When prefilling locally, the vllm scheduler will prioritize
+prefill request and pause any ongoing decode requests.
+Each decoder worker will launch its own prefill worker.
```
nnshah1, line 102: each decode worker launches its own prefill worker? I thought launching is done separately - will check rest

examples/python_rs/llm/vllm_nixl/README.md
```diff
@@ -35,42 +51,83 @@ All of the commands below are run inside the same container.
...
 ```
+ETCDCTL_API=3 etcdctl --endpoints=http://localhost:2379 put 'public/components/disagg_router/models/chat/deepseek-ai/DeepSeek-R1-Distill-Llama-8B' '{"max_local_prefill_length": <new_threshold>}'
+```
+
+### Workers
+
+```
+# start prefill worker in Terminal 1
+# for xPyD, please first start the decode workers
+# so that prefill workers can get metadata of all decode workers
```
nnshah1, line 122: no longer required - with etcd metadata discovery

examples/python_rs/llm/vllm_nixl/README.md
nnshah1: this seems off -I don't think the decode worker launches the prefill worker ... rather they are both launched as part of the deployment.

examples/python_rs/llm/vllm_nixl/README.md
```diff
@@ -80,19 +137,32 @@ CUDA_VISIBLE_DEVICES=1 python3 worker.py \
     '{"kv_connector":"TritonNixlConnector"}'
 ```
 
+Alternatively, we also provide a script to launch all workers in one go:
+```
+./start_single_node.sh
```
nnshah1, line 142: TODO = change to dynemo-deploy functionality

examples/python_rs/llm/vllm_nixl/prefill_worker.py
```diff
@@ -15,16 +15,19 @@
 
 
 import asyncio
+import os
 
-import msgspec
 import uvloop
-from common import find_remote_metadata, parse_vllm_args, temp_metadata_file
+from utils.nixl import find_remote_metadata, temp_metadata_file
```
nnshah1, line 21: needs merge to use etcd - from common

examples/python_rs/llm/vllm_nixl/prefill_worker.py
```diff
@@ -35,11 +38,7 @@ def __init__(self, engine_client):
         self.engine_client = engine_client
         print("RequestHandler initialized")
 
-    async def generate(self, raw_request: str):
-        request: RemotePrefillRequest = msgspec.json.decode(
-            raw_request.encode("utf-8"), type=RemotePrefillRequest
-        )
-
+    async def generate(self, request: RemotePrefillRequest):
```
nnshah1, line 41: nice!

examples/python_rs/llm/vllm_nixl/utils/nats_queue.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+class NATSQueue:
+    _instance: ClassVar[Optional["NATSQueue"]] = None
+    _lock: ClassVar[asyncio.Lock] = asyncio.Lock()
+
+    def __init__(
+        self,
+        stream_name: str = "default",
+        nats_server: str = "nats://localhost:4222",
```
nnshah1, line 33: can we read from environment variable - same as for rust ?

examples/python_rs/llm/vllm_nixl/utils/nats_queue.py
```diff
@@ -0,0 +1,137 @@
...
+    async def get_instance(
+        cls,
+        *,
+        stream_name: str = "default",
+        nats_server: str = "nats://localhost:4222",
+        dequeue_timeout: float = 1,
+    ):
+        """Get or create a singleton instance of NATSq"""
+        async with cls._lock:
+            if cls._instance is None:
```
nnshah1, line 56: still unsure what the lock buys us here - aren't we already protected by the GIL ? Could we create the singleton are the module level itself and hide the class? Something like:

```
class NatsQueue_


singleton = None
def NatsQueue(args):
     global sigleton
     if singleton: 
        return singleton
   ```

File: examples/python_rs/llm/vllm_nixl/utils/nats_queue.py [hunk 10]
```diff
@@ -0,0 +1,137 @@
...
+            try:
+                yield cls._instance
+            except Exception:
+                if cls._instance:
+                    await cls._instance.close()
+                cls._instance = None
+                raise
+
+    @classmethod
+    async def shutdown(cls):
```
nnshah1, line 72: seems redunant if we can do 

get_instance().close()

examples/python_rs/llm/vllm_nixl/utils/nats_queue.py
```diff
@@ -0,0 +1,137 @@
...
+    async def connect(self):
+        """Establish connection and create stream if needed"""
+        try:
+            if self.nc is None:
+                self.nc = NATS()
+                await self.nc.connect(self.nats_url)
+                self.js = self.nc.jetstream()
+                # Check if stream exists, if not create it
+                try:
+                    await self.js.stream_info(self.stream_name)
```
nnshah1, line 88: my understanding is that it is safe to add multiple times- may be simpler to always add - as long as the subject list is the same.

examples/python_rs/llm/vllm_nixl/utils/nats_queue.py
```diff
@@ -0,0 +1,137 @@
...
+                except NotFoundError:
+                    await self.js.add_stream(
+                        name=self.stream_name, subjects=[self.subject]
+                    )
+                    print(f"Stream '{self.stream_name}' created")
```
nnshah1, line 93: use logging - todo

examples/python_rs/llm/vllm_nixl/utils/nats_queue.py
```diff
@@ -0,0 +1,137 @@
...
+        """Close the connection when done"""
+        if self.nc:
+            await self.nc.close()
+            self.nc = None
+            self.js = None
+            self._subscriber = None
+
+    async def enqueue_task(self, task_data) -> None:
+        """
+        Enqueue a task using msgspec-encoded data
```
nnshah1, line 117: should we encode here then? Or type hint to bytes?

examples/python_rs/llm/vllm_nixl/utils/nats_queue.py
```diff
@@ -0,0 +1,137 @@
...
+
+    async def close(self):
+        """Close the connection when done"""
+        if self.nc:
+            await self.nc.close()
+            self.nc = None
+            self.js = None
+            self._subscriber = None
+
+    async def enqueue_task(self, task_data) -> None:
```
nnshah1, line 115: question is task or request better?

examples/python_rs/llm/vllm_nixl/utils/nats_queue.py
```diff
@@ -0,0 +1,137 @@
...
+        """
+        Enqueue a task using msgspec-encoded data
+        """
+        await self.ensure_connection()
+        try:
+            await self.js.publish(f"{self.stream_name}.queue", task_data)
```
nnshah1, line 121: can we make these private?

```
_js
_nc 
_stream_name?
```

examples/python_rs/llm/vllm_nixl/utils/nats_queue.py
```diff
@@ -0,0 +1,137 @@
...
+        except NatsError as e:
+            raise RuntimeError(f"Failed to enqueue task: {e}")
+
+    async def dequeue_task(self) -> Optional[bytes]:
```
nnshah1, line 125: should we go ahead and take a type here and handle msgpack details here?

examples/python_rs/llm/vllm_nixl/utils/prefill_queue.py
```diff
@@ -0,0 +1,56 @@
...
+        nats_server: str = "nats://localhost:4222",
+        dequeue_timeout: float = 1,
+    ):
+        super().__init__(
+            stream_name=stream_name,
+            nats_server=nats_server,
+            dequeue_timeout=dequeue_timeout,
+        )
+
+    async def enqueue_prefill_request(
```
nnshah1, line 42: I think we can use types in nats queue itself and then we can remove the need for class PrefillQueue entirely ... 

just take in an object to enqueue_task and to msgspec.json_encode() 

on dequeue - add a type paramter and auto decode in NatsQueue

examples/python_rs/llm/vllm_nixl/utils/vllm.py
```diff
@@ -0,0 +1,30 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
nnshah1, line 1: can we rename this to avoid clash with vllm package?

examples/python_rs/llm/vllm_nixl/worker.py
```diff
@@ -47,97 +46,142 @@ def __init__(
...
-        models = OpenAIServingModels(
-            engine_client=self.engine_client,
-            model_config=await self.engine_client.get_model_config(),
-            base_model_paths=[
-                BaseModelPath(
-                    name=self.model_name,
-                    model_path=self.model_name,
-                )
-            ],
+        self.disaggregated_router = disaggregated_router
```
nnshah1, line 59: do we want this as optional?

examples/python_rs/llm/vllm_nixl/worker.py
```diff
@@ -47,97 +46,142 @@ def __init__(
...
+        if do_remote_prefill:
+            assert (
+                disaggregated_router is not None
+            ), "Disaggregated router is required for remote prefill"
+
+        self.prefill_queue_nats_server = os.getenv(
+            "NATS_SERVER", "nats://localhost:4222"
```
nnshah1, line 66: can move to the NatsQueue

examples/python_rs/llm/vllm_nixl/worker.py
```diff
@@ -47,97 +46,142 @@ def __init__(
...
         )
-        self.openai_serving_chat = OpenAIServingChat(
-            engine_client=self.engine_client,
-            model_config=await self.engine_client.get_model_config(),
-            models=models,
-            request_logger=None,
-            response_role="assistant",
-            chat_template=None,
-            chat_template_content_format="auto",
+        self.prefill_queue_stream_name = model_name
```
nnshah1, line 68: can be private

examples/python_rs/llm/vllm_nixl/worker.py
```diff
@@ -47,97 +46,142 @@ def __init__(
...
-        self.initialized = True
+
+        print("RequestHandler initialized")
 
     def get_remote_prefill_request_callback(self):
+        # TODO: integrate prefill_queue to an triton_distributed endpoint
         async def callback(request: RemotePrefillRequest):
-            json_request = msgspec.json.encode(request).decode("utf-8")
-            self.prefill_client.generate(json_request)
+            async with PrefillQueue.get_instance(
```
nnshah1, line 78: can we pass a prefill queue instance in - and initialize once - in generate?

examples/python_rs/llm/vllm_nixl/worker.py
```diff
@@ -47,97 +46,142 @@ def __init__(
...
+    @triton_endpoint(vLLMGenerateRequest, MyRequestOutput)
     async def generate(self, request):
-        if not self.initialized:
-            await self.init()
-        assert self.openai_serving_chat is not None
-
-        if self.do_remote_prefill:
+        # TODO: consider prefix hit when deciding prefill locally or remotely
+        if self.do_remote_prefill and self.disaggregated_router.prefill_remote(
+            len(request.engine_prompt["prompt_token_ids"]), 0
```
nnshah1, line 90: decision is based on length alone?

examples/python_rs/llm/vllm_nixl/worker.py
```diff
@@ -47,97 +46,142 @@ def __init__(
...
-    @triton_endpoint(ChatCompletionRequest, ChatCompletionStreamResponse)
+    @triton_endpoint(vLLMGenerateRequest, MyRequestOutput)
     async def generate(self, request):
-        if not self.initialized:
-            await self.init()
-        assert self.openai_serving_chat is not None
-
-        if self.do_remote_prefill:
+        # TODO: consider prefix hit when deciding prefill locally or remotely
+        if self.do_remote_prefill and self.disaggregated_router.prefill_remote(
```
nnshah1, line 89: do we need both - could we do 
```
if self.disaggregatd_router is not None and ...
```

examples/python_rs/llm/vllm_nixl/worker.py
```diff
@@ -47,97 +46,142 @@ def __init__(
...
     prefill_client = (
-        await runtime.namespace("test-nixl")
+        await runtime.namespace("triton-init")
         .component("prefill")
         .endpoint("generate")
         .client()
     )
 
+    VLLM_WORKER_ID = endpoint.lease_id()
+    os.environ["VLLM_WORKER_ID"] = str(VLLM_WORKER_ID)
```
nnshah1, line 139: why do we need these environment variables? ... can we pass them as args?

examples/python_rs/llm/vllm_nixl/worker.py
```diff
@@ -47,97 +46,142 @@ def __init__(
         engine_client: EngineClient,
         prefill_client,
         do_remote_prefill: bool,
+        disaggregated_router: DisaggregatedRouter = None,
     ):
         self.model_name = model_name
-        self.engine_client = engine_client
+        self.client = engine_client
```
nnshah1, line 52: why switch from engine_client?

nnshah1 (commented):
Reviewd the python code and has some quesions /suggestions - think we can refactor to simplify

examples/python_rs/llm/vllm_nixl/prefill_worker.py
```diff
@@ -81,7 +78,28 @@ async def worker(runtime: DistributedRuntime, engine_args: AsyncEngineArgs):
             )
             for remote_metadata in remote_metadata:
                 await engine_client.add_remote_nixl_metadata(remote_metadata)
-            await endpoint.serve_endpoint(RequestHandler(engine_client).generate)
+
+            prefill_queue_nats_server = os.getenv(
```
nnshah1, line 82: would move this to nats queue itself - so done in one place for all queue users.

examples/python_rs/llm/vllm_nixl/prefill_worker.py
```diff
@@ -81,7 +78,28 @@ async def worker(runtime: DistributedRuntime, engine_args: AsyncEngineArgs):
...
+            # TODO: integrate prefill_queue to an triton_distributed endpoint
+            async with PrefillQueue.get_instance(
+                nats_server=prefill_queue_nats_server,
+                stream_name=prefill_queue_stream_name,
+            ) as prefill_queue:
+                while True:
+                    prefill_request = await prefill_queue.dequeue_prefill_request()
+                    if prefill_request is not None:
+                        vllm_logger.info(f"Dequeued prefill request: {prefill_request}")
+                        async for _ in RequestHandler(engine_client).generate(
```
nnshah1, line 99: do we need to a new request handler each time?

Or can we create once and just call generate multiple times?

examples/python_rs/llm/vllm_nixl/prefill_worker.py
```diff
@@ -61,11 +60,9 @@ async def generate(self, raw_request: str):
 
 @triton_worker()
 async def worker(runtime: DistributedRuntime, engine_args: AsyncEngineArgs):
-    component = runtime.namespace("test-nixl").component("prefill")
+    component = runtime.namespace("triton-init").component("prefill")
```
nnshah1, line 63: so technically this doesn't need access to the runtime?

As its getting and processing requests from the queue?

examples/python_rs/llm/vllm_nixl/README.md
tedzhouhk: nice catch, removed (don't know what I'm doing)

examples/python_rs/llm/vllm_nixl/utils/nats_queue.py
tedzhouhk: it's reading from env var in python worker script and passed here.

examples/python_rs/llm/vllm_nixl/utils/nats_queue.py
tedzhouhk: no expert in asyncio, I'm not sure if async will break GIL. If one thread is switched before the await and another thread kicks in, will this cause race condition? I'll add a TODO here to dig into this

tedzhouhk: also i heard recent python 3.12 make GIL optional :)

examples/python_rs/llm/vllm_nixl/utils/nats_queue.py
tedzhouhk: can we still do this way when get_instance() is async?

tedzhouhk: I'll add a TODO for this

examples/python_rs/llm/vllm_nixl/utils/nats_queue.py
tedzhouhk: yeah we can do that, it will just tell you the stream already exists. consider this code only runs once, i think it's fine to leave it like this

examples/python_rs/llm/vllm_nixl/utils/nats_queue.py
tedzhouhk: thanks, actually this is for debugging, I'll just remove it

examples/python_rs/llm/vllm_nixl/utils/nats_queue.py
tedzhouhk: added type hint

examples/python_rs/llm/vllm_nixl/utils/nats_queue.py
tedzhouhk: hmm, @ryanolson was suggesting having a general queue implementation and a prefill request wrapper. in fact `object` might be a better name.

tedzhouhk: i'll add a note

examples/python_rs/llm/vllm_nixl/utils/nats_queue.py
tedzhouhk: implemented as a general nats queue, I think bytes is better. User can create wrapper for other types.

examples/python_rs/llm/vllm_nixl/utils/prefill_queue.py
tedzhouhk: I'm OK for both

examples/python_rs/llm/vllm_nixl/utils/vllm.py
tedzhouhk: added a TODO, as I don't have a machine to test now  (don't want to break thing)

examples/python_rs/llm/vllm_nixl/worker.py
tedzhouhk: sorry, what do you mean?

examples/python_rs/llm/vllm_nixl/worker.py
tedzhouhk: yeah we can do that, but this will be replaced by a triton endpoint (as the todo suggest)

examples/python_rs/llm/vllm_nixl/worker.py
tedzhouhk: yeah this is the current version, we will replace it with more complex math model once vllm expose those and planner is there

examples/python_rs/llm/vllm_nixl/worker.py
```diff
@@ -47,97 +46,142 @@ def __init__(
...
+            prompt=request.engine_prompt,
+            sampling_params=request.sampling_params,
+            request_id=request.request_id,
             remote_prefill_params=remote_prefill_params,
         ):
-            if raw_response.startswith("data: [DONE]"):
-                break
-            response = json.loads(raw_response.lstrip("data: "))
-            yield response
+            yield MyRequestOutput(
```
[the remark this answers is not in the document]
tedzhouhk: this is inherited from the aggregated code, i think the reason is because vllm's request output is non-serializable (@alec-flowers to correct me)

examples/python_rs/llm/vllm_nixl/worker.py
tedzhouhk: no idea, inherited from aggregated code

tedzhouhk: will add a TODO

examples/python_rs/llm/vllm_nixl/worker.py
tedzhouhk: hmm, probably inherited from aggregated code

examples/python_rs/llm/vllm_nixl/prefill_worker.py
tedzhouhk: added a TODO

examples/python_rs/llm/vllm_nixl/prefill_worker.py
tedzhouhk: yeah, we can remove it. I'll add a TODO

tedzhouhk: but we will need this if the queue is integrated as part of the runtime

Issue #146, opened 2025-02-10
Add environment variables for tokio runtime thread counts


rmccorm4:
See https://github.com/triton-inference-server/triton_distributed/pull/129#discussion_r1948297305

TLDR; Instead of hard-coded thread counts 16/16 for tokio worker/blocking threads - we can set the default values to 16 and expose env vars like these to configure them:
```bash
TRITON_RUNTIME_NUM_WORKER_THREADS=16
TRITON_RUNTIME_MAX_BLOCKING_THREADS=16
```

Done criteria:
- Defaults are 16
- Env vars are exposed with a simple test to validate them
- Document env vars

PR #215, opened 2025-02-19
test: add unit tests for RuntimeConfig

tmonty12:
#### What does the PR do?

This PR adds unit tests to confirm env vars override defaults for worker and blocking threads count on `RuntimeConfig`.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID:

- closes GitHub issue: #146

### DIFF
--- a/runtime/rust/src/config.rs
+++ b/runtime/rust/src/config.rs
@@ -89,7 +89,7 @@ impl RuntimeConfig {
     /// Load the runtime configuration from the environment and configuration files
     /// Configuration is priorities in the following order, where the last has the lowest priority:
     /// 1. Environment variables (top priority)
-    /// TO DO: Add documentation for configuration files. Paths should be configurable.
+    ///     TO DO: Add documentation for configuration files. Paths should be configurable.
     /// 2. /opt/triton/etc/runtime.toml
     /// 3. /opt/triton/defaults/runtime.toml (lowest priority)
     ///
--- a/runtime/rust/src/config.rs
+++ b/runtime/rust/src/config.rs
@@ -35,7 +35,7 @@ impl WorkerConfig {
         // All calls should be global and thread safe.
         Figment::new()
             .merge(Serialized::defaults(Self::default()))
-            .merge(Env::prefixed("TRITON_WORKER_"))
+            .merge(Env::prefixed("TRD_WORKER_"))
             .extract()
             .unwrap() // safety: Called on startup, so panic is reasonable
     }
@@ -83,7 +83,14 @@ impl RuntimeConfig {
             .merge(Serialized::defaults(RuntimeConfig::default()))
             .merge(Toml::file("/opt/triton/defaults/runtime.toml"))
             .merge(Toml::file("/opt/triton/etc/runtime.toml"))
-            .merge(Env::prefixed("TRITON_RUNTIME_"))
+            .merge(Env::prefixed("TRD_RUNTIME_").filter_map(|k| {
+                let full_key = format!("TRD_RUNTIME_{}", k.as_str());
+                // filters out empty environment variables
+                match std::env::var(&full_key) {
+                    Ok(v) if !v.is_empty() => Some(k.into()),
+                    _ => None,
+                }
+            }))
     }
 
     /// Load the runtime configuration from the environment and configuration files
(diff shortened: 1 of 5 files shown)


runtime/rust/src/config.rs
```diff
@@ -161,3 +162,68 @@ pub fn jsonl_logging_enabled() -> bool {
...
+            },
+        )
+    }
+
+    #[test]
+    fn test_runtime_config_rejects_invalid_thread_count() -> Result<()> {
+        temp_env::with_vars(
+            vec![
+                ("TRITON_RUNTIME_NUM_WORKER_THREADS", Some("0")),
+                ("TRITON_RUNTIME_MAX_BLOCKING_THREADS", Some("0")),
```
rmccorm4, line 212: Not sure where the other touch points would be to make this change, but would be nice to keep a consistent prefix like `TRD_` instead of `TRITON_` in as many places as possible, so if we do a big find/replace later it can just be 1 or 2 instead of many. Same for `TRITON_WORKER_` above

Lmk what you think

rmccorm4: I think we have `TRD_*` in several places, `NOVA_*` in a couple, and `TRITON_*` here. But haven't done a grep yet to see.

tmonty12: Agreed we should have consistency of `TRD_` instead of `TRITON_`. The only other instance sof the use of `TRITON_` env var outside the scope of configuring the tokio worker threads are:

- https://github.com/triton-inference-server/triton_distributed/blob/31494da3568895302fb844f6939ffa39531542e2/runtime/rust/src/config.rs#L38

- https://github.com/triton-inference-server/triton_distributed/blob/31494da3568895302fb844f6939ffa39531542e2/runtime/rust/src/worker.rs#L108

I can update my PR to change `TRITON_` to `TRD_` in these places along with tokio worker thread config, including additional unit test coverage.

runtime/rust/src/config.rs
```diff
@@ -161,3 +162,68 @@ pub fn jsonl_logging_enabled() -> bool {
...
+            },
+        )
+    }
+
+    #[test]
+    fn test_runtime_config_defaults() -> Result<()> {
+        temp_env::with_vars(
+            vec![
+                ("TRITON_RUNTIME_NUM_WORKER_THREADS", None::<&str>),
+                ("TRITON_RUNTIME_MAX_BLOCKING_THREADS", None::<&str>),
```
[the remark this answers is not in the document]
tmonty12: I'll update my unit test to confirm `""` has the same behavior as `unset`

runtime/rust/src/config.rs
tmonty12: The `WorkerConfig` seems to be unused. It just defined the `graceful_shutdown_timeout` whereas the `RuntimeConfig` defines `num_worker_threads` and `max_blocking_threads`. Any reason these are separate? @ryanolson @rmccorm4

rmccorm4 (approved):

runtime/rust/src/config.rs
```diff
@@ -83,16 +83,24 @@ impl RuntimeConfig {
...
             .merge(Toml::file("/opt/triton/defaults/runtime.toml"))
             .merge(Toml::file("/opt/triton/etc/runtime.toml"))
-            .merge(Env::prefixed("TRITON_RUNTIME_"))
+            .merge(Env::prefixed("TRD_RUNTIME_").filter_map(|k| {
+                let full_key = format!("TRD_RUNTIME_{}", k.as_str());
+                // filters out empty environment variables
+                match std::env::var(&full_key) {
+                    Ok(v) if !v.is_empty() => Some(k.into()),
+                    _ => None,
+                }
```
rmccorm4, line 92: What happened when they were empty? Parsing errors?

tmonty12: Yeah you'd get the following error: `invalid type: found string "", expected usize for key "MAX_BLOCKING_THREADS" in `TRD_RUNTIME_` environment variable(s)`

grahamking (approved):

PR #217, opened 2025-02-20
feat: Add http support to vllm kv router example

alec-flowers:
#### What does the PR do?
- Adds support for http server for vLLM KV Routing
- Need to still add discover on ETCD to pass the MDC from worker to preprocessor, will do in future PR


examples/python_rs/llm/vllm/common/protocol.py
```diff
@@ -40,3 +44,70 @@ class Response(BaseModel):
...
+    lambda cls, source, handler: core_schema.any_schema()
+)
+
+
+class vLLMGenerateRequest(BaseModel):
+    """
+    Serializable class of all the fields vLLM engine requires for inference
+    """
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
ptarasiewiczNV, line 72: why is model config required?

alec-flowers: I don't need it in this situation. I think at one point I was unpacking a dict into it and wanted to allow arbitrary_types_allowed=True, but can drop for now.

ptarasiewiczNV (approved):

examples/python_rs/llm/vllm/common/protocol.py
ptarasiewiczNV: If it's redundant I would remove, but this is a minor thing, can be merged as is as well

alec-flowers: Removed.

ptarasiewiczNV (approved):

GuanLuo (approved):
LGTM, there may be conflict with my other PR just because they both modified the `worker.py` and `router.py`, but can be resolved when it happens.

nv-anants (approved):
readme changes look good

PR #225, opened 2025-02-20
build: Multi-stage VLLM Build

rmccorm4:
Overview
- Current dev image size is approaching 40GB as we keep appending to it, we need to start focusing on keeping the resulting image lean.
- This PR adds a PoC multi-stage build for Dockerfile.vllm
- This PR switches the base image from tritonserver to DLFW CUDA since we don't need the Triton specific bits right now

Before `38.6GB`:
```
$ docker images
REPOSITORY            TAG                        IMAGE ID       CREATED         SIZE
...
triton-distributed    v0.2.1.dev.31494da-vllm    e2accf223e4d   30 hours ago    38.6GB
```
After `19.7GB`:
```
$ docker images
REPOSITORY            TAG                        IMAGE ID       CREATED         SIZE
...
triton-distributed    v0.2.1.dev.728634d-vllm    94fd3db28a34   5 minutes ago   19.7GB
```

Follow-ups:
- We can add iterative improvements based on what we identify as the minimum set for the final runtime image
- We can explore better base images over time as needed based on requirements and target size
- Hopefully this can serve as a reference for some improvements to both STANDARD and TRTLLM images

Sanity Checks: 
- [x] Make sure VLLM KV routing vllm example works in leaner runtime image
- [x] Fix checks/CI


ptarasiewiczNV (dismissed):
This is awesome! Great change to `nvcr.io/nvidia/cuda-dl-base: 25.01-cuda12.8-devel-ubuntu24.04 ` as base. Also love the tmux scripts!

dmitrytokarev (approved):
From Docker build perspective high level things look good - building in dev container and copying artifacts into the final runtime container.
I don't see any significant wins to be made here as of now.
Already looking great! Awesome work! 👏 
Let's merge and then we can dissect each component to see if we copy and/or include anything that we don't want.

ishandhanani (approved):

PR #228, opened 2025-02-21
feat: pull-based prefill queue

tedzhouhk:
#### What does the PR do?

Implement a pull-based prefill queue over nats.

Note1: since there's not much logic besides calling the nats agent, I directly implement the queue in python instead of rust+python wrap.

Note2: this is a temporary solution, we would like to move the equeue operation after decode node allocate the kv blocks.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [x] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID:

- closes GitHub issue: #xxx

### DIFF
--- a/examples/python_rs/llm/vllm/common/prefill_queue.py
+++ b/examples/python_rs/llm/vllm/common/prefill_queue.py
@@ -1,0 +1,52 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from typing import Optional
+
+from common.protocol import PrefillRequest
+from triton_distributed_rs.nats_queue import NATSQueue
+
+
+class PrefillQueue(NATSQueue):
+    """
+    A wrapper of NATSQueue for PrefillRequest.
+    The stream name is forced to be "prefill_queue".
+    """
+
+    def __init__(
+        self,
+        stream_name="prefill_queue",
+        nats_server: str = "nats://localhost:4222",
+        dequeue_timeout: float = 1,
+    ):
+        super().__init__(
+            # force set stream_name to "prefill_queue"
+            stream_name="prefill_queue",
+            nats_server=nats_server,
+            dequeue_timeout=dequeue_timeout,
+        )
+
+    async def enqueue_prefill_request(self, prefill_request: PrefillRequest) -> None:
+        raw_request = prefill_request.model_dump_json()
+        await self.enqueue_task(raw_request)
+
+    async def dequeue_prefill_request(self) -> Optional[PrefillRequest]:
+        raw_request = await self.dequeue_task()
+        if raw_request is not None:
+            prefill_request = PrefillRequest.model_validate_json(raw_request)
+            return prefill_request
+        else:
+            return None
--- a/examples/python_rs/llm/vllm/disaggregated/decode_worker.py
+++ b/examples/python_rs/llm/vllm/disaggregated/decode_worker.py
@@ -15,6 +15,7 @@
 
 
 import asyncio
+import os
 import socket
 import uuid
 
@@ -27,8 +28,6 @@ from triton_distributed_rs import DistributedRuntime, triton_worker
 from triton_distributed_rs.prefill_queue import PrefillQueue
 from vllm.engine.arg_utils import AsyncEngineArgs
 from vllm.logger import logger as vllm_logger
-
-prefill_queue = PrefillQueue()
 
 
 class VllmDecodeEngine(BaseVllmEngine):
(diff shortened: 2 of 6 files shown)


tedzhouhk:
are we using `uv.lock` anywhere? I don't know how to update it.

ptarasiewiczNV:
I think it can be a general `RequestQueue` instead of `PrefillQueue`?

Can we hide it behind triton dist API?  Introduce `Client.queue` routing method, so instead of
```
self.prefill.generate(prefill_request) # or self.prefill.random(prefill_request)
```
We use:
```
self.prefill.queue(prefill_request)
```

python-wheel/python/triton_distributed_rs/nats_queue.py
```diff
@@ -0,0 +1,136 @@
...
+                self._subscriber = await self.js.pull_subscribe(
+                    f"{self.stream_name}.queue", durable="worker-group"
+                )
+        except NatsError as e:
+            await self.close()
+            raise ConnectionError(f"Failed to connect to NATS: {e}")
+
+    async def ensure_connection(self):
+        """Ensure we have an active connection"""
+        if self.nc is None or self.nc.is_closed:
```
nnshah1, line 103: is this for recovery? would it be possible to ensure this through creation, get instance() ?

python-wheel/python/triton_distributed_rs/nats_queue.py
```diff
@@ -0,0 +1,136 @@
...
+            try:
+                yield cls._instance
+            except Exception:
+                if cls._instance:
+                    await cls._instance.close()
+                cls._instance = None
+                raise
+
+    @classmethod
+    async def shutdown(cls):
```
nnshah1, line 71: why not use __del__ here and let the object shutdown on going out of scope?

python-wheel/python/triton_distributed_rs/nats_queue.py
```diff
@@ -0,0 +1,136 @@
...
+        """Close the connection when done"""
+        if self.nc:
+            await self.nc.close()
+            self.nc = None
+            self.js = None
+            self._subscriber = None
+
+    async def enqueue_task(self, task_data: str) -> None:
+        """Enqueue a task using JSON string from Pydantic model_dump_json()"""
+        await self.ensure_connection()
```
nnshah1, line 116: could we just use await connect() instead -and have connect() check if an existing connection is there?

python-wheel/python/triton_distributed_rs/nats_queue.py
```diff
@@ -0,0 +1,136 @@
...
+
+    async def close(self):
+        """Close the connection when done"""
+        if self.nc:
+            await self.nc.close()
+            self.nc = None
+            self.js = None
+            self._subscriber = None
+
+    async def enqueue_task(self, task_data: str) -> None:
```
nnshah1, line 114: could take in an object and convert to json internally - could take a pydantic object (check type) and do model_dump_json

python-wheel/python/triton_distributed_rs/nats_queue.py
```diff
@@ -0,0 +1,136 @@
...
+        """Enqueue a task using JSON string from Pydantic model_dump_json()"""
+        await self.ensure_connection()
+        try:
+            await self.js.publish(
+                f"{self.stream_name}.queue", task_data.encode("utf-8")
+            )
+        except NatsError as e:
+            raise RuntimeError(f"Failed to enqueue task: {e}")
+
+    async def dequeue_task(self) -> Optional[str]:
```
nnshah1, line 124: could offer an optional type argument and pass back a decoded task

examples/python_rs/llm/vllm/common/prefill_queue.py
```diff
@@ -0,0 +1,52 @@
...
+        dequeue_timeout: float = 1,
+    ):
+        super().__init__(
+            # force set stream_name to "prefill_queue"
+            stream_name="prefill_queue",
+            nats_server=nats_server,
+            dequeue_timeout=dequeue_timeout,
+        )
+
+    async def enqueue_prefill_request(self, prefill_request: PrefillRequest) -> None:
```
nnshah1, line 42: could simplify further and put model dump into parent.

Parent can take object typed as inheriting from base model

examples/python_rs/llm/vllm/common/prefill_queue.py
```diff
@@ -0,0 +1,52 @@
...
+        raw_request = prefill_request.model_dump_json()
+        await self.enqueue_task(raw_request)
+
+    async def dequeue_prefill_request(self) -> Optional[PrefillRequest]:
```
nnshah1, line 46: likewise could pass in a base model type and have parent handle the validation

examples/python_rs/llm/vllm/disaggregated/decode_worker.py
```diff
@@ -78,9 +80,10 @@ async def generate(self, raw_request):
             request_id=request_id,
         )
         vllm_logger.debug(f"Prefill request: {prefill_request}")
-        prefill_output = self.prefill.generate(
-            prefill_request.model_dump_json(),
-        )
+
+        # TODO: enqueue after the kv blocks are allocated
+        async with PrefillQueue.get_instance(nats_server=self.nats_server) as queue:
```
nnshah1, line 85: context manager makes me thing the queue gets iniatialized and is removed after the context is complete - but I think that is not the case here. 

Do we need a context manager?

examples/python_rs/llm/vllm/disaggregated/prefill_worker.py
```diff
@@ -71,8 +71,15 @@ async def worker(runtime: DistributedRuntime, engine_args: AsyncEngineArgs):
...
 
     async with VllmPrefillEngine(engine_args) as prefill_engine:
-        endpoint = component.endpoint("generate")
-        await endpoint.serve_endpoint(prefill_engine.generate)
+        nats_server = os.getenv("NATS_SERVER", "nats://localhost:4222")
+        # infinite loop to fetch prefill request from prefill queue
+        async with PrefillQueue.get_instance(nats_server=nats_server) as queue:
+            while True:
+                prefill_request = await queue.dequeue_prefill_request()
+                if prefill_request is not None:
```
nnshah1, line 79: what does the None case signify? should it exit the loop on None - or what conditions cause None? Should that be an assertion / exception case?

tedzhouhk:
close as it's part of https://github.com/triton-inference-server/triton_distributed/pull/209

PR #235, opened 2025-02-21
feat(tio): Distributed inference!

grahamking:
Add support in tio for distributed components and discovery.

Node 1:
```
tio in=http out=tdr://ns/backend/mistralrs
```

Node 2:
```
tio in=tdr://ns/backend/mistralrs out=mistralrs ~/llm_models/Llama-3.2-3B-Instruct
```

This will use etcd to auto-discover the model and NATS to talk to it. You can run multiple workers on the same endpoint and it will pick one at random each time.

The `ns/backend/mistralrs` are purely symbolic, pick anything as long as it has three parts, and it matches the other node.


llm/rust/triton-llm/src/http/service/discovery.rs
```diff
@@ -17,11 +17,9 @@ use std::sync::Arc;
 
 use serde::{Deserialize, Serialize};
 use tokio::sync::mpsc::Receiver;
-use tracing as log;
```
paulhendricks, line 20: 👍

applications/llm/tio/src/input/endpoint.rs
```diff
@@ -0,0 +1,86 @@
...
+            if elements.len() != 3 {
+                anyhow::bail!("An endpoint URL must have format {ENDPOINT_SCHEME}namespace/component/endpoint");
+            }
+
+            // Register with etcd
+            let endpoint = Endpoint {
+                namespace: elements[0].to_string(),
+                component: elements[1].to_string(),
+                name: elements[2].to_string(),
+            };
```
paulhendricks, line 47: This is a maybe good for another PR...but we should think about an `Endpoint.from_path(path: impl AsRef<str>)` or maybe `impl From<str> for Endpoint`.

applications/llm/tio/src/lib.rs
```diff
@@ -84,6 +101,33 @@ pub async fn run(
...
+            }
+            // This will attempt to connect to NATS and etcd
+            let distributed_runtime = DistributedRuntime::from_settings(runtime.clone()).await?;
+
+            let client = distributed_runtime
+                .namespace(elements[0])?
+                .component(elements[1])?
+                .endpoint(elements[2])
+                .client::<ChatCompletionRequest, Annotated<ChatCompletionResponseDelta>>()
+                .await?;
```
paulhendricks, line 117: Looks like it appears here as well? Maybe there's a way to do `DistributedRuntime.from_path()` that basically gets you to:

```rust
let client = distributed_runtime
                .namespace(elements[0])?
                .component(elements[1])?
                .endpoint(elements[2])
```

And does the validation as well.

paulhendricks (approved):

applications/llm/tio/src/input/endpoint.rs
grahamking: Agreed!

dmitrytokarev (approved):

dmitry-tokarev-nv (approved):
LGTM :shipit:

nnshah1:
@grahamking - starts to remind me of gst-launch :)

PR #239, opened 2025-02-22
ci: Add rust checks to missing directories

rmccorm4:
Changes:
- Parameterize rust directories and run the steps on all of them, rather than copying the steps for each directory
- Adds rust checks to some of the directories that were not previously having checks run
- Fix compiler warnings and formatting errors returned by the checks

Follow-ups:
- [ ] Most directories other than `lib/runtime` fail the license checks from `cargo-deny` - @dmitry-tokarev-nv I'll need help on fixing these errors, so I've skipped them for now to make sure everything else passes (They weren't being run before). Example failure: https://github.com/triton-inference-server/triton_distributed/actions/runs/13467873187/job/37637175712?pr=239
    - I think they may be missing a `deny.toml` 
- [ ] This could probably be further improved to only selectively trigger each based on changed paths, but didn't dive into this.


examples/rust/http/src/main.rs
```diff
@@ -69,7 +68,9 @@ async fn app(runtime: Runtime) -> Result<()> {
     // written to etcd
     // the cli when operating on an `http` component will validate the namespace.component is
     // registered with HttpServiceComponentDefinition
-    let component = distributed.namespace(&args.namespace)?.component(&args.component)?;
+    let component = distributed
+        .namespace(&args.namespace)?
+        .component(&args.component)?;
```
grahamking, line 73: Thanks for fixing these.

That means the `cargo fmt` check in CI is not working, right?

grahamking (dismissed):

grahamking:
I am very ambivalent about adding `cargo deny`. That will generate lots of busy work. The harder we make it for people to land code the less progress we'll make.

Is this a strict requirement from legal or something like that?

examples/rust/http/src/main.rs
rmccorm4: It was only being run on `runtime/rust` and `tio` directories, it was not being run on `examples/rust`, `llm/rust`, etc.

This PR adds the checks to the missing directories, and fixes the reported errors

.github/workflows/pre-merge-rust.yml
rmccorm4, line 1: From @grahamking:
> I am very ambivalent about adding cargo deny. That will generate lots of busy work. The harder we make it for people to land code the less progress we'll make.
>
> Is this a strict requirement from legal or something like that?

Yep this comes from legal/OSRB requirements - CC @nvda-mesharma @dmitry-tokarev-nv @nnshah1

nnshah1 (approved):

nv-anants (approved):

grahamking (approved):

examples/rust/http/src/main.rs
paulhendricks: I've noticed this in a few other PRs and my own. Where basically `cargo fmt` or the formatting isn't working and there's noise in the PRs with these formatting issues.

What are your guys thoughts on a pre-commit? That might help here.

PR #258, opened 2025-02-25
feat: Add HTTP completions endpoint to kv router example

sshchoi:
#### What does the PR do?
Add HTTP completions endpoint to the kv router python example

#### Commit Type:
- [x] feat

#### Test plan:

- CI Pipeline ID:


alec-flowers:
Need to run locally
`pre-commit run --show-diff-on-failure --color=always --all-files`

examples/python_rs/llm/vllm/kv_router/processor.py
```diff
@@ -124,13 +112,48 @@ async def generate(self, raw_request):
...
+                prompt_logprobs=output.prompt_logprobs,
+                outputs=output.outputs,
+                finished=output.finished,
+                metrics=output.metrics,
+            )
+
+            if request_type == RequestType.CHAT:
+                yield request_output
+            else:
+                yield (prompt_idx, request_output)
```
alec-flowers, line 145: Can you leave a comment why we need prompt_idx here. Not immediately clear to me.

examples/python_rs/llm/vllm/kv_router/processor.py
```diff
@@ -124,13 +112,48 @@ async def generate(self, raw_request):
...
+                prompt_token_ids=output.prompt_token_ids,
+                prompt_logprobs=output.prompt_logprobs,
+                outputs=output.outputs,
+                finished=output.finished,
+                metrics=output.metrics,
+            )
+
+            if request_type == RequestType.CHAT:
+                yield request_output
+            else:
```
alec-flowers, line 144: Can we do
```
elif request_type == RequestType.COMPLETION
    do this
else:
    raise NotImplementedError 
```

This is a public function so no guarantee user will pass what we expect.

examples/python_rs/llm/vllm/kv_router/processor.py
```diff
@@ -124,13 +112,48 @@ async def generate(self, raw_request):
...
-        output = self.generate_responses(engine_generator)
+        output = self.generate_responses(engine_generator, request_type)
 
         async for response in await self._stream_response(
             request, output, request_id, conversation
         ):
             yield response
 
+    async def generate_responses(
+        self, engine_generator, request_type
```
alec-flowers, line 123: type hints for args

examples/python_rs/llm/vllm/kv_router/processor.py
```diff
@@ -66,27 +74,7 @@ def _create_tokenizer(self, engine_args: AsyncEngineArgs) -> AnyTokenizer:
...
-                prompt_token_ids=output.prompt_token_ids,
-                prompt_logprobs=output.prompt_logprobs,
-                outputs=output.outputs,
-                finished=output.finished,
-                metrics=output.metrics,
-            )
-
-    @triton_endpoint(ChatCompletionRequest, ChatCompletionStreamResponse)
-    async def generate(self, raw_request):
+    async def _generate(self, raw_request, request_type: RequestType):
```
alec-flowers, line 77: nit: type hint for raw_request

examples/python_rs/llm/vllm/kv_router/processor.py
alec-flowers: Maybe we should also make function private as well (underscore in front)

examples/python_rs/llm/vllm/kv_router/processor.py
```diff
@@ -154,11 +177,17 @@ async def worker(runtime: DistributedRuntime, engine_args: AsyncEngineArgs):
 
     preprocess_component = runtime.namespace("triton-init").component("process")
     await preprocess_component.create_service()
-    preprocess_endpoint = preprocess_component.endpoint("chat/completions")
+
+    chat_endpoint = preprocess_component.endpoint("chat/completions")
+    completions_endpoint = preprocess_component.endpoint("completions")
 
     processor = Processor(engine_args, router_client, workers_client)
     assert isinstance(processor, ProcessMixIn)
```
alec-flowers, line 185: We can drop this now. I think I had added this before Processor inherited from the ProcessMixIn protocol. This should be already checked as part of the inheritence.

examples/python_rs/llm/vllm/common/chat_processor.py
```diff
@@ -33,31 +36,45 @@
...
     chat_processor: "ChatProcessor | None"
+    completions_processor: "CompletionsProcessor | None"
     model_config: ModelConfig
 
     def __init__(self):
         pass
 
+    def _get_processor(self, raw_request: dict):
+        # Determine the processor type based on the request structure
+        if hasattr(raw_request, 'messages'):
```
alec-flowers, line 59: If raw_request is truly a dict (unsure about this) then hasattr() should fail.

hasattr() is used to check if a class has a certain attribute. 
This would fail I think - hasattr(request_dict, 'messages')
But this would work - hasattr(request_dict, 'keys') because keys is a dict attribute.

I would change this to be
'messages' in raw_request

examples/python_rs/llm/vllm/common/chat_processor.py
```diff
@@ -33,31 +36,45 @@
...
 
     engine_args: AsyncEngineArgs
     chat_processor: "ChatProcessor | None"
+    completions_processor: "CompletionsProcessor | None"
     model_config: ModelConfig
 
     def __init__(self):
         pass
 
+    def _get_processor(self, raw_request: dict):
```
alec-flowers, line 57: We are doing an implicit check in this function, but would be nice to be explicit if we can. It shouldn't be hard to pass request_type down to this so we could combine being explicit by passing the request_type and then falling back in case we don't have it. 

def _get_processor(self, raw_request: dict, request_type: Type = None):
 if request_type is not None:
        if request_type == "chat":
            processor = self.chat_processor
        elif request_type == "completion":
            processor = self.completions_processor
        else:
            raise ValueError(f"Unknown request_type: {request_type}")

    elif isinstance(raw_request, dict) and 'messages' in raw_request:
        processor = self.chat_processor
    else:
        processor = self.completions_processor

examples/python_rs/llm/vllm/common/chat_processor.py
```diff
@@ -33,31 +36,45 @@
...
-            request_prompt,
-            engine_prompt,
-        ) = await self.chat_processor.preprocess(raw_request)
+        processor = self._get_processor(raw_request)
+        if processor is None:
+            raise RuntimeError("Processor has not been initialized")
+        request = processor.parse_raw_request(raw_request)
+        preprocess_result = await processor.preprocess(raw_request)
+
+        if isinstance(processor, ChatProcessor):
```
alec-flowers, line 71: If we standardize on a consistent output we could drop this if statement. 

```
class PreprocessResult(NamedTuple):
    request_prompt: dict
    engine_prompt: dict
    conversation: Optional[Any] = None
```

Potentially something like this and have the processor.preprocess() return this tuple

and then can return at the end

    return request, result.conversation, result.request_prompt, result.engine_prompt, sampling_params

examples/python_rs/llm/vllm/common/chat_processor.py
```diff
@@ -146,3 +163,66 @@ async def stream_response(
...
+        self.tokenizer = tokenizer
+        self.model_config = model_config
+        self.openai_serving = OpenAIServingCompletion(
+            engine_client=None,
+            model_config=model_config,
+            models=None,
+            request_logger=None,
+        )
+
+    def parse_raw_request(self, raw_request: dict) -> CompletionRequest:
```
alec-flowers, line 178: This dict type hint is wrong, otherwise `raw_request.model` syntax will fail.

examples/python_rs/llm/vllm/common/chat_processor.py
```diff
@@ -146,3 +163,66 @@ async def stream_response(
...
+            model=raw_request.model,
+            prompt=raw_request.prompt,
+            max_tokens=raw_request.max_tokens,
+            temperature=raw_request.temperature,
+            top_p=raw_request.top_p,
+            top_k=raw_request.top_k,
+            stream=raw_request.stream,
+        )
+
+    async def preprocess(self, raw_request: dict) -> tuple[ConversationMessage, RequestPrompt, TokensPrompt]:
```
alec-flowers, line 189: nit: return type is wrong, 3 elements specified but only 2 returned.

examples/python_rs/llm/vllm/common/chat_processor.py
```diff
@@ -146,3 +163,66 @@ async def stream_response(
...
+
+    async def stream_response(
+        self,
+        request: CompletionRequest,
+        result_generator: AsyncIterator,
+        request_id: str,
+        conversation: None,
+    ):
+        request_metadata = RequestResponseMetadata(request_id=request_id)
+        assert request.stream, "Only stream is supported"
```
alec-flowers, line 213: Lets swap assertions with 
```
if not request.stream:
    raise ValueError("Only streaming responses are supported")
```

examples/python_rs/llm/vllm/common/chat_processor.py
```diff
@@ -146,3 +163,66 @@ async def stream_response(
...
+        )
+
+        return request_prompts[0], engine_prompts[0]
+
+    async def stream_response(
+        self,
+        request: CompletionRequest,
+        result_generator: AsyncIterator,
+        request_id: str,
+        conversation: None,
```
alec-flowers, line 210: This type is odd think you were going for ?
`conversation: Optional[str] = None,`

alec-flowers (approved):
Left a few comments. Once resolved should be good to merge.

examples/python_rs/llm/vllm/common/chat_processor.py
alec-flowers: Looking at it again I am thinking the type is wrong for raw_request.

examples/python_rs/llm/vllm/kv_router/processor.py
```diff
@@ -129,13 +124,55 @@ async def generate(self, raw_request):
...
+                # For chat requests, yield the request_output directly.
+                yield request_output
+            elif request_type == RequestType.COMPLETION:
+                # Completion requests can have multiple prompts and stream generator requires the prompt index
+                yield (prompt_idx, request_output)
+            else:
+                raise NotImplementedError(
+                    f"Request type {request_type} not implemented"
+                )
+        prompt_idx += 1
```
rmccorm4, line 164: Is this `prompt_idx` used / correct? Is it supposed to be incremented outside of the loop? I don't think it would have persistent state across function calls here, so seems like it would always be zero.

rmccorm4 (commented):
Generally LGTM but one question on `prompt_idx`

alec-flowers (dismissed):
LGTM

examples/python_rs/llm/vllm/kv_router/processor.py
sshchoi: prompt_idx should always be 0 because we only support taking in one prompt per request. I will get rid of the line that increments prompt_idx as it doesn't do anything here

rmccorm4 (approved):

examples/python_rs/llm/vllm/common/chat_processor.py
```diff
@@ -94,12 +128,12 @@ def __init__(self, tokenizer: AnyTokenizer, model_config: ModelConfig):
             chat_template_content_format="auto",
         )
 
-    def parse_raw_request(self, raw_request: dict) -> ChatCompletionRequest:
+    def parse_raw_request(
+        self, raw_request: ChatCompletionRequest
+    ) -> ChatCompletionRequest:
```
GuanLuo, line 133: Parsing from and to the same type? i.e. `ChatCompletionRequest`

PR #265, opened 2025-02-25
feat: tio support preprocessor

grahamking:
Add backend type `EngineConfig::StaticCore` that wraps the engine in a preprocessor (prompt templating and tokenization).

Add example engine `echo_core` (`out=echo_core`) which takes and returns tokens. A nice side effect is that it echos the full prompt template with system prompt, whereas `echo_full` echos only user prompt.

image

### DIFF
--- a/applications/llm/tio/src/input/http.rs
+++ b/applications/llm/tio/src/input/http.rs
@@ -15,8 +15,19 @@
 
 use std::sync::Arc;
 
-use triton_distributed_llm::http::service::{discovery, service_v2};
-use triton_distributed_runtime::{DistributedRuntime, Runtime};
+use triton_distributed_llm::{
+    backend::Backend,
+    http::service::{discovery, service_v2},
+    preprocessor::OpenAIPreprocessor,
+    types::{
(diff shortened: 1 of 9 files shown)


applications/llm/tio/src/input/http.rs
```diff
@@ -55,6 +66,32 @@ pub async fn run(
...
         }
+        EngineConfig::StaticCore {
+            service_name,
+            engine: inner_engine,
+            card,
+        } => {
+            let frontend = ServiceFrontend::<
+                SingleIn<ChatCompletionRequest>,
+                ManyOut<Annotated<ChatCompletionResponseDelta>>,
+            >::new();
```
paulhendricks, line 77: Question: are we able to type alias this? `<SingleIn<ChatCompletionRequest>, ManyOut<Annotated<ChatCompletionResponseDelta>>`

Or is there a good reason we don't want to do that?

It just feels like that is such an established pattern we're using at this point that if we can, it might be worth it just to reduce the signature complexity.

Maybe like `ChatStreamingIO`? Not suggesting that for this PR - just thoughts.

paulhendricks (approved):

paulhendricks (approved):

PR #270, opened 2025-02-25
refactor: endpoint -> function

ryanolson:
#### What does the PR do?

- renaming endpoint -> function for describing the logical pieces

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID:

- closes GitHub issue: #xxx

### DIFF
--- a/examples/python_rs/llm/vllm/disaggregated/decode_worker.py
+++ b/examples/python_rs/llm/vllm/disaggregated/decode_worker.py
@@ -40,7 +40,7 @@ from triton_distributed.runtime import (
 
 class VllmDecodeEngine(BaseVllmEngine, ProcessMixIn):
     """
-    Request handler for the generate endpoint
+    Request handler for the generate function
     """
 
     def __init__(self, engine_args: AsyncEngineArgs, prefill):
@@ -110,7 +110,7 @@ class VllmDecodeEngine(BaseVllmEngine, ProcessMixIn):
 @triton_worker()
 async def worker(runtime: DistributedRuntime, engine_args: AsyncEngineArgs):
     """
-    Instantiate a `backend` component and serve the `generate` endpoint
+    Instantiate a `backend` component and serve the `generate` function
     A `Component` can serve multiple endpoints
     """
     component = runtime.namespace("triton-init").component("vllm")
@@ -119,12 +119,12 @@ async def worker(runtime: DistributedRuntime, engine_args: AsyncEngineArgs):
     prefill = (
         await runtime.namespace("triton-init")
         .component("prefill")
-        .endpoint("generate")
+        .function("generate")
         .client()
     )
     async with VllmDecodeEngine(engine_args, prefill) as decode_engine:
-        endpoint = component.endpoint("generate")
-        await endpoint.serve_endpoint(decode_engine.generate)
+        function = component.function("generate")
+        await function.serve_endpoint(decode_engine.generate)
 
 
 if __name__ == "__main__":
--- a/examples/python_rs/llm/vllm/disaggregated/prefill_worker.py
+++ b/examples/python_rs/llm/vllm/disaggregated/prefill_worker.py
@@ -33,7 +33,7 @@ from triton_distributed.runtime import (
 
 class VllmPrefillEngine(BaseVllmEngine):
     """
-    Request handler for the generate endpoint
+    Request handler for the generate function
     """
 
     def __init__(self, engine_args: AsyncEngineArgs):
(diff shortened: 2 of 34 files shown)


lib/runtime/src/component.rs
```diff
@@ -187,127 +191,10 @@ impl ComponentBuilder {
...
-
-    pub fn name(&self) -> &str {
-        &self.name
-    }
-}
-
 // Custom validator function
 fn validate_allowed_chars(input: &str) -> Result<(), ValidationError> {
     // Define the allowed character set using a regex
-    let regex = regex::Regex::new(r"^[a-z0-9-_]+$").unwrap();
```
nnshah1, line 310: does this need to change?

examples/python_rs/llm/vllm/disaggregated/decode_worker.py
```diff
@@ -119,12 +119,12 @@ async def worker(runtime: DistributedRuntime, engine_args: AsyncEngineArgs):
...
         .component("prefill")
-        .endpoint("generate")
+        .function("generate")
         .client()
     )
     async with VllmDecodeEngine(engine_args, prefill) as decode_engine:
-        endpoint = component.endpoint("generate")
-        await endpoint.serve_endpoint(decode_engine.generate)
+        function = component.function("generate")
+        await function.serve_endpoint(decode_engine.generate)
```
nnshah1, line 127:
```suggestion
        await function.serve_function(decode_engine.generate)
```

examples/python_rs/llm/vllm/disaggregated/prefill_worker.py
```diff
@@ -69,15 +69,15 @@ async def generate(self, request):
...
     A `Component` can serve multiple endpoints
     """
     component = runtime.namespace("triton-init").component("prefill")
     await component.create_service()
 
     async with VllmPrefillEngine(engine_args) as prefill_engine:
-        endpoint = component.endpoint("generate")
-        await endpoint.serve_endpoint(prefill_engine.generate)
+        function = component.function("generate")
+        await function.serve_endpoint(prefill_engine.generate)
```
nnshah1, line 80:
```suggestion
        await function.serve_function(prefill_engine.generate)
```

lib/runtime/src/component.rs
ryanolson: yes, it was incorrect.

examples/python_rs/llm/vllm/disaggregated/decode_worker.py
ryanolson: it should be `serve_on_endpoint` or `serve_as_endpoint` or `serve`

Another later feature would be to `create_endpoint` which would return a handle to the task which was spawned.

examples/python_rs/llm/vllm/kv_router/processor.py
```diff
@@ -141,25 +141,25 @@ async def generate(self, raw_request):
...
     Set up clients to the router and workers.
-    Serve the triton-init.process.chat/completions endpoint.
+    Serve the triton-init.process.chat/completions function.
     """
     workers_client = (
         await runtime.namespace("triton-init")
         .component("vllm")
-        .endpoint("generate")
+        .function("generate")
         .client()
```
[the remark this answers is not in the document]
ryanolson: Currently, `create_service()` and `serve_endpoint()` do not have any options, but in the future, there are arguments that will provide ways to tune/customize these objects and their behaviors.

lib/runtime/src/component.rs
ryanolson: The name should begin with [a-z] and end with [a-z0-9] with `-` and `_` only allowed in between.

examples/python_rs/llm/vllm/disaggregated/decode_worker.py
nnshah1: Recommend "serve". 

What if we made that a function on component?
```
component.serve(decode_engine.generate)
```

If we keep the endpoint terminology - I think actually the previous makes sense:

```
endpoint = component.endpoint("generate")
endpoint.serve_function(decode_engine.generate)
```

lib/runtime/src/component.rs
nnshah1: maybe a comment to describe the allowed pattern via few examples?

PR #277, opened 2025-02-26
feat: Endpoint defaults for namespace/component/other

grahamking:
This means we don't need to explain the parts to the users until they are ready. We use what they provide and default the rest.

Allows all of this and more:
- `tio out=tdr://test`
- `tio out=tdr://llama_8b_pool`
- `tio in=tdr://corp_ai_research_group/model_next-20250226`
- `tio out=tdr://AIRE.NIM.migrate.mistralrs.1802`

Python, API, etc all untouched.

### DIFF
--- a/lib/runtime/src/protocols.rs
+++ b/lib/runtime/src/protocols.rs
@@ -14,7 +14,6 @@
 // limitations under the License.
 
 use serde::{Deserialize, Serialize};
-use std::convert::TryFrom;
 use std::str::FromStr;
 
 use crate::pipeline::PipelineError;
@@ -22,6 +21,15 @@ use crate::pipeline::PipelineError;
 pub mod annotated;
 
 pub type LeaseId = i64;
+
+/// Default namespace if user does not provide one
+const DEFAULT_NS: &str = "NS";
+
+/// Default component name
+const DEFAULT_C: &str = "C";
+
+/// Default endpoint/function name
+const DEFAULT_E: &str = "E";
 
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
 pub struct Component {
@@ -31,7 +39,7 @@ pub struct Component {
 
 /// Represents an endpoint with a namespace, component, and name.
 ///
-/// An `Endpoint` is defined by a three-part string separated by `/`:
+/// An `Endpoint` is defined by a three-part string separated by `/` or a '.':
 /// - **namespace**
 /// - **component**
 /// - **name**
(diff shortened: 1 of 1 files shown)


lib/runtime/src/protocols.rs
```diff
@@ -23,6 +22,15 @@ pub mod annotated;
 
 pub type LeaseId = i64;
 
+/// Default namespace if user does not provide one
+const DEFAULT_NS: &str = "NS";
```
nnshah1, line 26: what if we spelled them out, 
"DEFAULT_NAMESPACE"
"DEFAULT_COMPONENT"
"DEFAULT_ENDPOINT"?

lib/runtime/src/protocols.rs
```diff
@@ -31,57 +39,103 @@ pub struct Component {
 
 /// Represents an endpoint with a namespace, component, and name.
 ///
-/// An `Endpoint` is defined by a three-part string separated by `/`:
+/// An `Endpoint` is defined by a three-part string separated by `/` or a '.':
```
nnshah1, line 42: do we want to make this less flexible to allow dots in path segments?

Or could we adopt uri validation - follow what ever is valid for a uri?

nnshah1: actually we should definitely validate as a uri - since the syntax is that - people will assume its a valid uri - we can fail if things are provided that don't make sense -

lib/runtime/src/protocols.rs
```diff
@@ -31,57 +39,103 @@ pub struct Component {
...
+    pub namespace: String,
+    pub component: String,
     pub name: String,
+}
 
-    /// Component of the endpoint.
-    pub component: String,
+impl PartialEq<Vec<&str>> for Endpoint {
+    fn eq(&self, other: &Vec<&str>) -> bool {
+        if other.len() != 3 {
```
nnshah1, line 57: can we make '3' a constant ?

lib/runtime/src/protocols.rs
```diff
@@ -31,57 +39,103 @@ pub struct Component {
...
+            return false;
+        }
 
-    /// Namespace of the component.
-    pub namespace: String,
+        self.namespace == other[0] && self.component == other[1] && self.name == other[2]
```
nnshah1, line 61: maybe constants for the fields, "NAMESPACE_INDEX", "COMPONENT_INDEX", "NAME_INDEX"

lib/runtime/src/protocols.rs
```diff
@@ -31,57 +39,103 @@ pub struct Component {
...
+    }
+}
+
+impl PartialEq<Endpoint> for Vec<&str> {
+    fn eq(&self, other: &Endpoint) -> bool {
+        other == self
```
nnshah1, line 67: nice!

lib/runtime/src/protocols.rs
```diff
@@ -31,57 +39,103 @@ pub struct Component {
...
+impl From<&str> for Endpoint {
+    /// Creates an `Endpoint` from a string.
     ///
     /// # Arguments
     /// - `path`: A string in the format `"namespace/component/endpoint"`.
     ///
-    /// # Errors
-    /// Returns a `PipelineError::InvalidFormat` if the input string does not
-    /// have exactly three parts separated by `/`.
+    /// The first two parts become the first two elements of the vector.
```
nnshah1, line 87: nice!

lib/runtime/src/protocols.rs
grahamking: Good point, doing.

lib/runtime/src/protocols.rs
grahamking: The spec always said "namespace.component.endpoint", I was surprised we also did slash. The slash matches etcd, and natural paths, so it makes sense. So I guess both for the moment.

lib/runtime/src/protocols.rs
grahamking: Here and below (`other[0]`) can't change unless everything else changes, so it's not something we need to be able to adjust in a constant.

lib/runtime/src/protocols.rs
nnshah1: are they both valid uris?

lib/runtime/src/protocols.rs
```diff
@@ -31,57 +37,103 @@ pub struct Component {
...
     /// ```
-    fn try_from(path: &str) -> Result<Self, Self::Error> {
-        let elements: Vec<&str> = path.split('/').collect();
-        if elements.len() != 3 {
-            return Err(PipelineError::InvalidEndpointFormat);
+    fn from(input: &str) -> Self {
+        let mut result = Endpoint::default();
+
+        // Split the input string on either '.' or '/'
+        let elements: Vec<&str> = input
```
nnshah1, line 109: can we use a uri parser?

lib/runtime/src/protocols.rs
```diff
@@ -90,10 +142,10 @@ impl FromStr for Endpoint {
 
     /// Parses an `Endpoint` from a string using the standard Rust `.parse::<T>()` pattern.
     ///
-    /// This is implemented in terms of [`TryFrom<&str>`].
+    /// This is implemented in terms of [`From<&str>`].
     ///
     /// # Errors
-    /// Returns an `PipelineError::InvalidFormat` if the input does not match `"namespace/component/endpoint"`.
+    /// Does not fail
```
nnshah1, line 148: do we need to validate the scheme?

lib/runtime/src/protocols.rs
```diff
@@ -229,86 +238,63 @@ mod tests {
...
     }
 
     #[test]
-    fn test_invalid_endpoint_try_from() {
-        let input = "invalid_endpoint_format";
-
-        let result = Endpoint::try_from(input);
-        assert!(result.is_err(), "Parsing should fail for an invalid format");
+    fn test_endpoint_from() {
+        let result = Endpoint::from("component");
```
nnshah1, line 250: question: If you provide only one value  - I thought it was the namespace - , 

but is the rule if one value it is -> component
if two values it is namespace/component
if three it is namepsace/component/x_[others]

?

lib/runtime/src/protocols.rs
grahamking: Probably. I'd like to wait until the endpoint / function / etc discussion resolves before making the change.

lib/runtime/src/protocols.rs
```diff
@@ -31,57 +37,103 @@ pub struct Component {
...
+        let elements: Vec<&str> = input
+            .trim_matches([' ', '/', '.'])
+            .split(['.', '/'])
+            .filter(|x| !x.is_empty())
+            .collect();
+
+        match elements.len() {
+            0 => {}
+            1 => {
+                result.component = elements[0].to_string();
```
nnshah1, line 118: got it!

lib/runtime/src/protocols.rs
nnshah1: got it from above

nnshah1 (approved):
LGTM - would have slightly more verbose default values than NS, C, E

Also - do we need a default 'C" - maybe they have to provide 'C' and NS and E can be default.

lib/runtime/src/protocols.rs
grahamking: They are URLs, so a bit stricter that URI. We have our own scheme (the `tdr` in `tdr://`), anything after that is up to us. In the [usual](https://datatracker.ietf.org/doc/html/rfc3986#section-3) syntax yes both would be valid. The dot separator would parse as the authority (domain name in http) and the slash separated would parse as the path.

The path being a URL is not fixed so I don't want to over-commit at this stage, it just felt natural for a user. I don't know where we're headed with the networked components.

lib/runtime/src/protocols.rs
grahamking: Yes! It will be invisible to the user, that's just what felt most natural to me.

grahamking:
> LGTM - would have slightly more verbose default values than NS, C, E
> 
> Also - do we need a default 'C" - maybe they have to provide 'C' and NS and E can be default.

Ideally I'd use the project name as the default namespace, but, euh, you see the problem!

paulhendricks (approved):
Looks good, great work! 🚢 

Now that we have what is looking like a well defined `Endpoint`, does it make sense to move it to it's own `endpoint.rs` file?

Not suggesting for this PR but instead a follow up.

grahamking:
> does it make sense to move it to it's own `endpoint.rs` file?

Yes, except it takes a lot of effort to land PRs right now, you get one or two a day at most. I need to save those for big changes. Hopefully we'll smooth out the speed bumps with time and get to the smaller code-quality work.

PR #280, opened 2025-02-26
fix: Fix stream::until_deadline bug and improve metric examples

rmccorm4:
#### What does the PR do?

- Fix bug in `stream::until_deadline` that caused the call to `scrape_stats` to block forever. 
    - Essentially, the `deadline` was never getting reached because there was no async task in the stream to initiate the deadline check on the call to `poll_next`. 
    - By changing this to a `tokio::async::sleep_until` to place an async sleep task back on the stream, it would wake on the deadline and properly exit.
- Use `clap` CLI args instead of `ENV` vars for component/endpoint for clearer instructions and error messages.
- Make polling interval for count configurable with `--poll-interval`
- Add example of a custom `stats_handler` to the `service_metrics` server example to show how you could see custom stats outputs on calls to `component.scrape_stats` in both `count` and the `service_metrics` client example.
- Add some additional debug logging and comments for clarity on certain areas


applications/llm/count/src/main.rs
```diff
@@ -74,27 +92,27 @@ pub struct LLMWorkerLoadCapacity {
 
 fn main() -> Result<()> {
     logging::init();
+    let args = Args::parse();
     let worker = Worker::from_settings()?;
-    worker.execute(app)
+    worker.execute(|runtime| app(runtime, args))
```
grahamking, line 97: You could move `let args = Args::parse()` into the `app` function.

grahamking (approved):

PR #282, opened 2025-02-26
refactor: service/endpoint stats_handler

ryanolson:
#### What does the PR do?

- simplifies adding a stats handler to an endpoint

#### Checklist
- [x] Related issues are referenced.

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID:

- closes GitHub issue: #xxx


lib/runtime/src/component.rs
```diff
@@ -73,9 +74,15 @@ pub enum TransportType {
...
+#[derive(Default)]
+pub struct RegistryInner {
+    services: HashMap<String, Service>,
+    stats_handlers: HashMap<String, Arc<std::sync::Mutex<HashMap<String, EndpointStatsHandler>>>>,
+}
+
 #[derive(Clone)]
 pub struct Registry {
-    services: Arc<tokio::sync::Mutex<HashMap<String, Service>>>,
+    inner: Arc<tokio::sync::Mutex<RegistryInner>>,
```
grahamking, line 85: A mutex of a map of mutexes. Living la vida loca.

lib/runtime/src/component/endpoint.rs
```diff
@@ -33,34 +33,62 @@ pub struct EndpointConfig {
...
 
-        // let group = service.group(service_name.as_str());
+        // get the stats handler map
+        let handler_map = registry
+            .stats_handlers
+            .get(&service_name)
+            .cloned()
+            .expect("no stats handler registry; this is unexpected");
+
+        drop(registry);
```
grahamking, line 83: The clone and drop outer lock is nice. stats_handlers is string->arc so it's a small clone, and we don't have to worry about lock acquisition order.

lib/runtime/src/component/service.rs
```diff
@@ -32,31 +40,31 @@ pub struct ServiceConfig {
...
-        let mut guard = component.drt.component_registry.services.lock().await;
+        let stats_handler_registry: Arc<Mutex<HashMap<String, EndpointStatsHandler>>> =
+            Arc::new(Mutex::new(HashMap::new()));
 
-        if guard.contains_key(&component.etcd_path()) {
+        let stats_handler_registry_clone = stats_handler_registry.clone();
+
+        let mut guard = component.drt.component_registry.inner.lock().await;
+
+        if guard.services.contains_key(&service_name) {
```
grahamking, line 67: Could you acquire and use the lock on one line (65 + 67)? Or `drop(guard)`. Then we're not holding it for the duration of the `secondary.spawn` below. That would make me more comfortable, it's easier to reason about.

lib/runtime/src/component/service.rs
```diff
@@ -65,23 +73,36 @@ impl ServiceConfigBuilder {
...
             })
             .await?
             .map_err(|e| anyhow::anyhow!("Failed to start service: {e}"))?;
 
-        guard.insert(component.etcd_path(), service);
+        // new copy of service_name as the previous one is moved into the task above
+        let service_name = component.service_name();
+
+        // insert the service into the registry
+        guard.services.insert(service_name.clone(), service);
```
grahamking, line 98: ... and then re-acquire the lock here

grahamking (approved):
I'd like to see the lock drop change I commented on, but that can also go in later.

lib/runtime/src/component.rs
ryanolson: Indeed. The first mutex protects the creation step. 

The handler map is an arc, so a clone of that is given to the service's stats handler. 

When the stats handler is called, only only one mutex acquisition is required. 

To add a new endpoint with a new handler, both mutexes need to be acquired, but it's a very very low frequency event. Usually just at startup.

ryanolson:
Gotcha. 

Yes, I agree. I'm will do a minor refactor. 

- drop the registry lock
- do any fallable setup for the async tasks before launching it. 
- if error, clean up the entry in the registry. 
- if not, do a final registry update.

ryanolson:
@grahamking - removed the spawn altogether. locking, starting, if start fails the lock is dropped on the throw.  the lock is preventing multiple calling from registering a component service with the same name in the same lease. There can be only one.

If for some reason we were to allow multiples, they would just need to be created with separate leases.  For now, you can only host one copy of a component as a service per instance.

grahamking (approved):

lib/runtime/src/component/service.rs
```diff
@@ -32,56 +40,64 @@ pub struct ServiceConfig {
...
+        tracing::debug!("Starting service: {}", service_name);
+        let service = builder
+            .description(description)
+            .stats_handler(move |name, stats| {
+                log::trace!("stats_handler: {name}, {stats:?}");
+                let mut guard = stats_handler_registry.lock().unwrap();
+                match guard.get_mut(&name) {
+                    Some(handler) => handler(stats),
+                    None => serde_json::Value::Null,
+                }
```
GuanLuo, line 83: So there will be a default stats handler for the service for completeness. And whenever an endpoint is created, I can provide a endpoint specific stats handler?

PR #288, opened 2025-02-27
DO NOT MERGE: Update decode_worker.py

piotrm-nvidia:


rmccorm4 (changes requested):
Can mark as draft to avoid merge and avoid triggering reviews

piotrm-nvidia:
I'm sorry for any inconvenience.

The purpose of this PR is to test CI flow for real PR not draft. The CI is configured to push to Gitlab only changes touching some files assigned to reviewers so I can't create such PR without spamming your.

piotrm-nvidia:
I also can't remove you from this PR so spamming complains should be directed to author of rules, which prevent me from doing what I need to do. Why I can't share with single person some PR before I will spam everyone?

piotrm-nvidia:
I just merged CI modification and it is the only way to check if it is fine. Do PR here and see in gitlab if all things pass.

The sandbox will be nice for sure but we also need the single source of true and it is the CI.

PR #298, opened 2025-02-27
feat: llama.cpp engine for tio

grahamking:
Docs in README


paulhendricks (approved):
Looks good - great work! 

tgi next? ;)

PR #300, opened 2025-02-27
ci: use commit message to pass ci variables

nv-anants:


piotrm-nvidia (approved):

nvda-mesharma:
JET pipeline failed?

nv-anants:
> JET pipeline failed?

its failed cause the vllm build did not run. Its a CI fix, we need to enable vllm if Jet is true

nv-anants:
> > Enable passing gitlab ci variable via commit message.
> 
> What's the use case for this? I've never heard of someone trying that before.

There have been quite a few cases where (long running) tests were broken due to incoming PRs. This mainly enables us to run on demand tests form the ci. We cannot run them for every PR due to resource limitations and running it automatically based on file changes and such is also not possible cause we do not know what changes will break them. So this provides an easy option for PR owners to run extra set of tests, in case they think PR will impact non standard tests.

.github/workflows/trigger_ci.py
```diff
@@ -0,0 +1,163 @@
...
+def extract_options_from_commit(commit_message):
+    """
+    Extract CI options from commit message.
+    Supports two formats:
+    1. Boolean flags: [option]
+    2. Valued options: [option=value]
+    """
+    options = {}
+    # Look for text between square brackets
+    for part in commit_message.split("["):
```
saturley-hall, line 57: I am fine with this but my personal commit style is sometimes `fix[ci]: environment variable name` which I think would trigger the warning for an unsupported option below. If we are enumerating the options and checking against them why not just iterate a RegEx over the keys in the supported map?

.github/workflows/trigger_ci.yml
```diff
@@ -50,29 +54,37 @@ jobs:
             - 'examples/python_rs/llm/**'
```
saturley-hall, line 54: Should we be filtering out changes to the `trigger-ci.py` above?

.github/workflows/trigger_ci.py
```diff
@@ -0,0 +1,163 @@
...
+
+import requests
+
+ALLOWED_CI_OPTIONS = {
+    # Boolean flags
+    "RUN_JET_TESTS": {"type": "bool"},
+    "NIGHTLY_BENCHMARK": {"type": "bool"},
+    "RUN_VLLM": {"type": "bool"},
+    # Options with values
+    "CI_DEFAULT_BRANCH": {"type": "str"},  # placeholder for future use
```
saturley-hall, line 27: Just as a security thought, could I mount an attack here? Is there somewhere we wrap/filter the string?

saturley-hall (approved):
I think this looks fine for a first pass but there are a couple areas that we might want to exert some thought toward.

PR #308, opened 2025-02-28
feat: vllm engine

grahamking:
triton-distributed-llm component and support in tio

### DIFF
--- a/launch/tio/README.md
+++ b/launch/tio/README.md
@@ -106,3 +106,36 @@ The extra `--model-config` flag is because:
 - We don't yet read it out of the GGUF (TODO), so we need an HF repo with `tokenizer.json` et al
 
 If the build step also builds llama_cpp libraries into `target/release` ("libllama.so", "libggml.so", "libggml-base.so", "libggml-cpu.so", "libggml-cuda.so"), then `tio` will need to find those at runtime. Set `LD_LIBRARY_PATH`, and be sure to deploy them alongside the `tio` binary.
+
+## vllm
+
+Using the [vllm](https://github.com/vllm-project/vllm) Python library. We only use the back half of vllm, talking to it over `zmq`. Slow startup, fast inference. Supports both safetensors from HF and GGUF files.
+
+We use [uv](https://docs.astral.sh/uv/) but any virtualenv manager should work.
+
+Setup:
+```
+uv venv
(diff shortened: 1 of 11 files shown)


launch/tio/README.md
```diff
@@ -106,3 +106,36 @@ The extra `--model-config` flag is because:
...
+source .venv/bin/activate
+uv pip install pip
+uv pip install vllm setuptools
+```
+
+**Note: If you're on Ubuntu 22.04 or earlier, you will need to add `--python=python3.10` to your `uv venv` command**
+
+Build:
+```
+cargo build --release --features vllm
```
rmccorm4, line 128: Quick question - why do mistralrs and llamacpp also require `--features cuda` but not `vllm`? Is it because they're written directly in rust and involve touching the memory management, whereas `vllm` is being talked to out-of-process over ZMQ + python, doing its own h2d<->d2h internally, and is CPU in<->out with Rust process otherwise?

paulhendricks (approved):

launch/tio/src/main.rs
```diff
@@ -83,6 +83,28 @@ fn main() -> anyhow::Result<()> {
                 panic!("Rebuild with --features=sglang");
             }
         }
+
+        #[allow(unused_variables)]
+        if flags.internal_vllm_process {
```
rmccorm4, line 88: Out of curiosity - can launching the vllm subprocess be a part of starting the vllm worker, and remove the need for this internal flag? Or can you clarify why this internal flag approach is preferred?

rmccorm4 (approved):
LGTM, but left a couple questions

launch/tio/README.md
grahamking: I'm not sure. I've been assuming that it's because vllm and sglang do the detection at runtime.

launch/tio/src/main.rs
grahamking: After the fork/clone we're doing an `execve` with `tio --internal-vllm-process` as the binary to run. tio's subprocess is itself with that flag. A user would never run that directly. (We're not actually doing fork/execve, `Command::spawn` wraps it up).

I guess we could just `clone` and continue with the subprocess part in the new process. Is that what you're thinking?

Issue #306, opened 2025-02-28
vllm.monolith import fails at CI


piotrm-nvidia:
https://gitlab-master.nvidia.com/dl/triton/triton-distributed/-/jobs/145688580#L9190

```
                f"{formatted_tb}"
            ) from e
E           _pytest.nodes.Collector.CollectError: ImportError while importing test module '/workspace/examples/python_rs/llm/vllm/monolith/__init__.py'.
E           Hint: make sure your test modules/packages have valid Python names.
E           Traceback:
E           /root/.local/share/uv/python/cpython-3.12.9-linux-x86_64-gnu/lib/python3.12/importlib/__init__.py:90: in import_module
E               return _bootstrap._gcd_import(name[level:], package, level)
E           E   ModuleNotFoundError: No module named 'vllm.monolith'
config     = <_pytest.config.Config object at 0x7f369733f6b0>
exc_info   = <ExceptionInfo ModuleNotFoundError("No module named 'vllm.monolith'") tblen=1>
exc_repr   = ExceptionChainRepr(reprtraceback=ReprTraceback(reprentries=[ReprEntry(lines=['    return _bootstrap._gcd_import(name[l.../python3.12/importlib/__init__.py', lineno=90, message="ModuleNotFoundError: No module named 'vllm.monolith'"), None)])
formatted_tb = "/root/.local/share/uv/python/cpython-3.12.9-linux-x86_64-gnu/lib/python3.12/importlib/__init__.py:90: in import_modul... return _bootstrap._gcd_import(name[level:], package, level)\nE   ModuleNotFoundError: No module named 'vllm.monolith'"
importmode = 'prepend'
path       = PosixPath('/workspace/examples/python_rs/llm/vllm/monolith/__init__.py')
/opt/triton/venv/lib/python3.12/site-packages/_pytest/python.py:523: CollectError
------------ generated xml file: /workspace/vllm_pytest_report.xml -------------
```

piotrm-nvidia:
The PR 307 moved failure to different file:

https://gitlab-master.nvidia.com/dl/triton/triton-distributed/-/jobs/145745502/viewer#L194

```
config     = <_pytest.config.Config object at 0x7f9a3abc1550>
exc_info   = <ExceptionInfo ModuleNotFoundError("No module named 'vllm.common'") tblen=1>
[...]

PR #315, opened 2025-02-28
Revert "Add missing inits (#307)"

piotrm-nvidia:


piotrm-nvidia:
See: https://github.com/triton-inference-server/triton_distributed/issues/306

The change doesn't improve NIXL builds so let's revert it. It is neutral for non vLLM code.

PR #318, opened 2025-03-01
[fix] OpenAI API object: completion to text_completion

alec-flowers:
#### What does the PR do?
Follow openAI spec.

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test


rmccorm4 (approved):
There may be some tests that fail, and there's one right below here `create_test_delta` that hard-codes object="completion" - lets change that one too

rmccorm4 (approved):

